PR #123, opened 2020-06-02
show results from chinese_db_crnn_server

MissPenguin:


GreatV:
Thank you for reporting this issue. We'll look into it!

PR #236, opened 2020-06-19
If img is None then .shape method will raise an exception 

xxxpsyduck:


xxxpsyduck:
whoops it seems like no need to change in db_process.py

PR #366, opened 2020-07-17
add the missing improt for east_process.py and format some codes

BeyondYourself:
add the missing improt for east_process.py and format some codes


ppocr/data/det/east_process.py
```diff
@@ -16,7 +16,7 @@
 import cv2
 import numpy as np
 import json
-
+import sys
```
littletomatodonkey, line 19: 把这个地方的报错信息优化一下吧
https://github.com/PaddlePaddle/PaddleOCR/blob/455769b13a38eaf97fb77736fd327528e61575ff/ppocr/data/det/east_process.py#L509

PR #472, opened 2020-08-03
add a file path separator

BeyondYourself:


ppocr/data/det/east_process.py
```diff
@@ -52,7 +52,7 @@ def convert_label_infor(self, label_infor):
         label_infor = label_infor.decode()
         label_infor = label_infor.encode('utf-8').decode('utf-8-sig')
         substr = label_infor.strip("\n").split("\t")
-        img_path = self.img_set_dir + substr[0]
+        img_path = self.img_set_dir + "/" + substr[0]
```
littletomatodonkey, line 55: The function `os.path.join()` is a preference.

ppocr/utils/save_load.py
```diff
@@ -109,9 +109,14 @@ def init_model(config, program, exe):
...
-        path = checkpoints
-        fluid.load(program, path, exe)
-        logger.info("Finish initing model from {}".format(path))
+        if os.path.exists(checkpoints + '.pdparams'):
+            path = checkpoints
+            fluid.load(program, path, exe)
+            logger.info("Finish initing model from {}".format(path))
+        else:
+            raise ValueError(
+                "Model checkpoints  {} does not exists,"
```
littletomatodonkey, line 118: `checkpoints  {}` -> `checkpoints {}` (delete one of the two blanks)
`prefix .` -> `prefix.`

ppocr/data/det/east_process.py
BeyondYourself: Done
there is case :  path1=os.path.join("aaa/bbb/","ccc") ,the result is :"aaa/bbb/","ccc"
 path2 = os.path.join("aaa/bbb/","/ccc")   ,the result is :"ccc"
path3 = os.path.join("aaa/bbb","/ccc")   ,the result is :"ccc"
so the image path in the label could not be start with a file separator

ppocr/utils/save_load.py
BeyondYourself: Done, this pr is in [https://github.com/PaddlePaddle/PaddleOCR/pull/467](url)

PR #625, opened 2020-08-27
fix some bugs in dataset_traversal

BeyondYourself:
fix the fix read and remove some unused import;

improve the label read, it dose not need to split based on the ' \t' and remove the spaceblank at left/right before we split


BeyondYourself:
@littletomatodonkey  please review this pr

ppocr/data/rec/dataset_traversal.py
```diff
@@ -268,7 +264,7 @@ def sample_iter_reader():
                             infer_mode=True)
                     yield norm_img
             else:
-                with open(self.label_file_path, "rb") as fin:
```
littletomatodonkey, line 271: 这块改回 `with open(xxx, "rb")`吧

BeyondYourself: OK

BeyondYourself: 此处还是建议使用encoding,后续使用就避免出现忘记decode操作

PR #637, opened 2020-08-31
add the img path to the function process_image

BeyondYourself:
The purpose of adding image path is to distinguish which image has the problem when we process the images


BeyondYourself:
@tink2123 please review this pr,thx

BeyondYourself:
@littletomatodonkey  please review this pr ,thx

ppocr/data/rec/img_tools.py
```diff
@@ -339,6 +339,7 @@ def warp(img, ang):
 
 
 def process_image(img,
+                  img_path,
```
littletomatodonkey, line 342: 把添加的img path参数写在最后把，同时指定`img_path=""`，防止影响其他的函数，上面的参数顺序也调整下

BeyondYourself: OK

PR #682, opened 2020-09-07
fix non-square image random crop width and height does not correspond

lyl120117:
非正方形图片使用随机裁剪有BUG


ppocr/data/det/random_crop_data.py
```diff
@@ -124,20 +124,21 @@ def RandomCropData(data, size):
     # 计算crop区域
     crop_x, crop_y, crop_w, crop_h = crop_area(im, all_care_polys,
                                                min_crop_side_ratio, max_tries)
+    dh, dw = size
     # crop 图片 保持比例填充
```
LDOUBLEV, line 128: 顺便把这些中文注释删掉吧

LDOUBLEV:
as: https://github.com/PaddlePaddle/PaddleOCR/pull/600

PR #757, opened 2020-09-19
fix whl doc and add multi-language

littletomatodonkey:
as title.


WenmuZhou:
good job to big old

LDOUBLEV (approved):
great job to big old

PR #848, opened 2020-09-25
add comments for rec

tink2123:


ppocr/utils/character.py
```diff
@@ -132,6 +151,21 @@ def cal_predicts_accuracy(char_ops,
                           labels,
                           labels_lod,
                           is_remove_duplicate=False):
+    """
+    Calculate predicts accrarcy
```
littletomatodonkey, line 155: prediction

tools/program.py
```diff
@@ -150,19 +150,20 @@ def check_gpu(use_gpu):
...
     """
     Build a program using a model and an optimizer
-        1. create feeds
-        2. create a dataloader
-        3. create a model
-        4. create fetchs
-        5. create an optimizer
+        1. create a dataloader
+        2. create a model
+        3. create fetchs
```
littletomatodonkey, line 155: fetches

ppocr/utils/character.py
tink2123: Done, thx

tools/program.py
tink2123: Done, thx

ppocr/utils/character.py
tink2123: done, thx

tools/program.py
tink2123: Done, thx

PR #947, opened 2020-10-16
Fix the table view in deploy/lite/readme_en.md

Airine:
Missing spaces cause the anomaly display of the table in section 4.


littletomatodonkey (approved):
LGTM

PR #1045, opened 2020-10-28
add the character_dict_path to the yml

BeyondYourself:
if the character type is not en, the param character_dict_path is necessary  when we use other dict


littletomatodonkey (approved):
LGTM

PR #1263, opened 2020-11-27
remove unused info and fix some bugs

BeyondYourself:
update readme and fix some bugs 
eg:For a image without any label information , the file status will change when you click the Save button


Evezerest:
Except for the import part, do not delete any code of this file especially those not commented, which are (or will be) useful

BeyondYourself:
> Except for the import part, do not delete any code of this file especially those not commented, which are (or will be) useful

Only remove redundant commented code，it is s not recommended for users.These codes don't have any userful  Introduction information

BeyondYourself:
@grasswolfs PTAL

PR #1299, opened 2020-12-01
add predict_cls to predict_system

WenmuZhou:


tools/infer/predict_system.py
```diff
@@ -88,6 +92,15 @@ def __call__(self, img):
             tmp_box = copy.deepcopy(dt_boxes[bno])
             img_crop = self.get_rotate_crop_image(ori_im, tmp_box)
             img_crop_list.append(img_crop)
+            cv2.imwrite(
+                '/home/zhoujun20/dygraph/PaddleOCR_rc/inference_results/{}.jpg'.
```
littletomatodonkey, line 96: delete it?

tools/infer/predict_system.py
```diff
@@ -88,6 +92,15 @@ def __call__(self, img):
...
+                format(bno), img_crop)
+        if self.use_angle_cls:
+            img_crop_list, angle_list, elapse = self.text_classifier(
+                img_crop_list)
+            print("cls num  : {}, elapse : {}".format(
```
littletomatodonkey, line 101: print the tme unit at the sam time?

littletomatodonkey, line 101: use logger info?

tools/infer/predict_system.py
```diff
@@ -149,8 +159,13 @@ def main(args):
...
 
-            draw_img = draw_ocr(
-                image, boxes, txts, scores, drop_score=drop_score)
+            draw_img = draw_ocr_box_txt(
+                image,
+                boxes,
+                txts,
+                scores,
+                drop_score=drop_score,
+                font_path=font_path)
```
littletomatodonkey, line 168: make it configurable?

tools/infer/predict_system.py
WenmuZhou: Only need to print the whole time is enough

PR #1344, opened 2020-12-07
paddleocr whl adaptation dygraph

WenmuZhou:
The update is as follows

1. paddleocr supports input of network pictures or numpy arrays
2. paddleocr whl support dygraph
3. drop_score will affect the output, not the drawing
4. Support the use of grayscale image as input, which will be converted to BGR in paddleocr


LDOUBLEV (approved):
lgtm

PR #1350, opened 2020-12-08
Fix path in doc of quantization

YukSing12:
Fix incorrect path of Mobile terminal model deployment in doc of quantization.


LDOUBLEV (approved):
lgtm

PR #1377, opened 2020-12-10
Fix bug

YukSing12:
Move https://github.com/PaddlePaddle/PaddleOCR/blob/23942cf7bce004eb9152fd191a8a5e0d13c9e513/tools/export_model.py#L18
after https://github.com/PaddlePaddle/PaddleOCR/blob/23942cf7bce004eb9152fd191a8a5e0d13c9e513/tools/export_model.py#L24
to avoid ModuleNotFoundError: No module named 'ppocr'

### DIFF
--- a/tools/export_model.py
+++ b/tools/export_model.py
@@ -15,7 +15,7 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


LDOUBLEV (approved):
LGTM

PR #1397, opened 2020-12-12
[Dygraph] add model list doc, update model link in paddeocr.py and update model size in quick start doc

WenmuZhou:


littletomatodonkey (approved):
LGTM

PR #1399, opened 2020-12-12
PPOCRLabel update TopSpeed Model

edencfc:
1. Directly add label when draw a box
2. Directly double click to change a label


Evezerest:
reviewing

PR #1599, opened 2020-12-29
add CyclicalCosineDecay

littletomatodonkey:
1. Add CyclicalCosineDecay. The strategy can be seen in : https://arxiv.org/pdf/2012.12645.pdf. In ocr, without any fine tune on hyper-parameters, it can directly takes a 0.3% accuracy gain for CRNN(with 10% more training epochs).
2. modify the epoch num, because if we want to save models per 5epochs and total epoch is set as 50, then the 50th epoch  model will not be saved for the code before(the last epoch is actually 49).

### DIFF
--- a/ppocr/optimizer/lr_scheduler.py
+++ b/ppocr/optimizer/lr_scheduler.py
@@ -1,0 +1,49 @@
+# copyright (c) 2020 PaddlePaddle Authors. All Rights Reserve.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


ppocr/optimizer/lr_scheduler.py
```diff
@@ -0,0 +1,49 @@
+# copyright (c) 2020 PaddlePaddle Authors. All Rights Reserve.
```
WenmuZhou, line 1: 这个文件不需要吧

littletomatodonkey: 类似于框架中的lr_scheduler.py，和ocr中封装的learning_rate分开，感觉会更清楚一点

LDOUBLEV (approved):

WenmuZhou (approved):

PR #1861, opened 2021-01-28
fixed wrong naming of German model

Justus-Jonas:
The german recognition model was falsely named French recognition


LDOUBLEV:
Thanks for your contribution

LDOUBLEV (approved):

PR #2052, opened 2021-02-20
IMPROVE Project Readability

weijiang2009:
=> Some docs are designed to be read by Chinese but not so 'International'

=>Let's improve the English Docs Readability for the project


weijiang2009:
CLOSE due to switching to the new git dev branch

PR #2055, opened 2021-02-21
improve style text infer process

littletomatodonkey:
att, improve style text infer process when the text is too long.

for the corpus: `Paddle12345678901234littletomatodonkey`

1. methods before: repeat bg according to the text input.(Cut off may also happen when is text is too long and the effect is not good for this case)

image

2. split text according to the style input, generate and then combine the results.

image


weisy11:
Good job! LGTM

LDOUBLEV (approved):

PR #2150, opened 2021-03-01
bengali dict and corpus added

dynamicguy:
Please change the short code for Bengali language from bg to bn. That's the ISO code.


dynamicguy:
CLA signing not working. I've already signed the CLA but it won't show up in checks. here is a [screenshot](https://www.dropbox.com/s/gwfr2ybe6vd3fis/Screenshot%20from%202021-04-01%2003-32-14.png?dl=0)

PR #2278, opened 2021-03-17
Modify a self-assignment in rec_postprocess.py

ShixunWu:
In 'def add_special_char(self, dict_character)', 
I saw that 'dict_character = dict_character'. 
Shouldn't it be  'self.dict_character = dict_character'?


LDOUBLEV (approved):

ppocr/postprocess/rec_postprocess.py
```diff
@@ -139,9 +139,9 @@ def __init__(self,
     def add_special_char(self, dict_character):
         self.beg_str = "sos"
         self.end_str = "eos"
-        dict_character = dict_character
```
WenmuZhou, line 142: 一种更好的做法是直接删掉这一行

PR #2562, opened 2021-04-20
fix loader cause it does not need to break

littletomatodonkey:
att, the following is the log after the change, which is same as before between 2 epochs. 

image

### DIFF
--- a/tools/program.py
+++ b/tools/program.py
@@ -190,16 +190,12 @@ def train(config,
         start_epoch = 1
 
     for epoch in range(start_epoch, epoch_num + 1):
-        train_dataloader = build_dataloader(
-            config, 'Train', device, logger, seed=epoch)
         train_batch_cost = 0.0
         train_reader_cost = 0.0
         batch_sum = 0
         batch_start = time.time()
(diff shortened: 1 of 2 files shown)


tools/program.py
```diff
@@ -190,16 +190,12 @@ def train(config,
         start_epoch = 1
 
     for epoch in range(start_epoch, epoch_num + 1):
-        train_dataloader = build_dataloader(
```
WenmuZhou, line 193: 这个保留，去掉了之后每个epoch内的数据内容就是一样的了，之前的动态采样

littletomatodonkey: `DistributedBatchSampler`在这里会做shuffle，每次重新loader()的时候都会调用一次，

https://github.com/PaddlePaddle/Paddle/blob/f0cc188399214d35755269ed98076ffbdc69d586/python/paddle/fluid/dataloader/batch_sampler.py#L266

littletomatodonkey: done，已修改~感谢

JetHong (approved):

WenmuZhou (approved):

PR #3112, opened 2021-06-16
add embedding and fix eval when distillation

littletomatodonkey:
att

### DIFF
--- a/tools/eval.py
+++ b/tools/eval.py
@@ -44,8 +44,15 @@ def main():
     # build model
     # for rec algorithm
     if hasattr(post_process_class, 'character'):
-        config['Architecture']["Head"]['out_channels'] = len(
-            getattr(post_process_class, 'character'))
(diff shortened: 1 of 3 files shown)


tools/eval.py
```diff
@@ -44,8 +44,15 @@ def main():
     # build model
     # for rec algorithm
     if hasattr(post_process_class, 'character'):
-        config['Architecture']["Head"]['out_channels'] = len(
-            getattr(post_process_class, 'character'))
+        char_num = len(getattr(post_process_class, 'character'))
+        if config['Architecture']["algorithm"] in ["Distillation",
```
LDOUBLEV, line 48: ["Distillation",] : # distillation mode
逗号和注释去掉吧

littletomatodonkey: 方便扩展，

LDOUBLEV (approved):

PR #3227, opened 2021-06-30
Fixed Imghdr missing some cases

mohamadmansourX:
`imghdr.what` method fail in checking some JPG images.
`imghdr.what` use the following command to check for JPG by checking the header: `[6:10] in (b'JFIF', b'Exif')`
However, its not always the case as some - quantized JPG - can start with `b'\xff\xd8\xff\xdb' `headers and not including `(b'JFIF', b'Exif')` as described in the references [here](https://www.digicamsoft.com/itu/itu-t81-36.html) and [here](https://web.archive.org/web/20120403212223/http://class.ee.iastate.edu/ee528/Reading%20material/JPEG_File_Format.pdf).
`\xdb` defining the Quantization Table.

PIL can be an alternative for that issue until imghdr bug is fixed [issue](https://github.com/python/cpython/pull/26964#issue-680660918)

### DIFF
--- a/ppocr/utils/utility.py
+++ b/ppocr/utils/utility.py
@@ -16,7 +16,7 @@ import logging
 import os
 import imghdr
 import cv2
-
+from PIL import Image
 
 def print_dict(d, logger, delimiter=0):
     """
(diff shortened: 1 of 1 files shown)


mohamadmansourX:
The [bug](https://github.com/python/cpython/pull/26964#issue-680660918) was fixed.

PR #3988, opened 2021-09-09
Gradio Web Demo

AK391:
easy to use web demo using gradio on huggingface spaces


AK391:
@tink2123 please see the PR thanks

MissPenguin:
Thanks for your contribution! But the demo seems not work for now. Could u please check it out?

PR #4645, opened 2021-11-13
feat: add support to process https

### LINKED ISSUE #4644: `.ocr()` does not support loading images with https
While adding paddleOCR to [Jina Hub](https://hub.jina.ai/executor/78yp7etm), I find that `.ocr()` does not support `https`. Here are sample codes which do not work.

```python
ocr = PaddleOCR(use_angle_cls=True, lang='en', use_gpu=False)
img_path = 'https://files.realpython.com/media/results2.5d8d3b8108cd.png'
result = ocr.ocr(img_path, cls=True)
for line in result:
    print(line)
```

nan-wang:
- resolve #4644 with the python built-in [`urllib.parse()`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse)
- fix the misleading logging message

### DIFF
--- a/paddleocr.py
+++ b/paddleocr.py
@@ -22,6 +22,7 @@ import cv2
 import logging
 import numpy as np
 from pathlib import Path
+from urllib.parse import urlparse
 
 from tools.infer import predict_system
 from ppocr.utils.logging import get_logger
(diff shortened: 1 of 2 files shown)


nan-wang:
@tink2123 I think I need your approval for running the ci.

tink2123:
Thanks for your contribution. PaddleOCR has CI to check the model format and correctness. In order to pass CI, you need to add `urllib` in requirements.txt.

nan-wang:
> Thanks for your contribution. PaddleOCR has CI to check the model format and correctness. In order to pass CI, you need to add `urllib` in requirements.txt.

I don't think this is case because `urllib` is in the standard python library. https://docs.python.org/3/library/urllib.html

PR #4676, opened 2021-11-17
rename multiheadAttention.py

techkang:
Old filename `multiheadAttention` is camelCase and doesn't meet pep8 standard, it should be lower_case.


littletomatodonkey (commented):
thanks for your contribution, you should also rename the file at the same time.

techkang:
The file is already renamed.

PR #4748, opened 2021-11-24
pair param with key when load trained model params

WenmuZhou:


ppocr/utils/save_load.py
```diff
@@ -56,9 +56,26 @@ def load_model(config, model, optimizer=None):
     if checkpoints:
         if checkpoints.endswith('pdparams'):
```
littletomatodonkey, line 57: 修改为.pdparams吧

WenmuZhou: done tks

littletomatodonkey (approved):

PR #4807, opened 2021-12-01
Create Turkish Dict

aliosmankaya:


tink2123:
Thanks for contributing dictionaries and corpus. In order to avoid the repo is too large, we will not join in temporarily. Your corpus will be used in subsequent iterations.

The characters in your dictionary have been covered in paddleocr Latin model. If you are interested, you can test the effect of the model.

```
pip install paddelocr>=2.3.0
paddleocr --lang=tr --image_dir=xxx
```

aliosmankaya:
Thank you! I can't wait for the Turkish model.

PR #4905, opened 2021-12-13
fix bug in copypaste when point num > 8

WenmuZhou:


littletomatodonkey (approved):

ppocr/data/simple_dataset.py
```diff
@@ -93,7 +94,8 @@ def get_ext_data(self):
                 img = f.read()
                 data['image'] = img
             data = transform(data, load_data_ops)
-            if data is None:
+
+            if data is None or data['polys'].shape[1]!=4:
```
LDOUBLEV, line 98: 需要安装
pre-commit

PR #4917, opened 2021-12-14
Update pipeline_http_client.py

muare:
the second test_img_dir assignment is not necessary:
after it is defined,  it's not modified. We can use it anyway.


MissPenguin (approved):

MissPenguin:
LGTM

PR #5175, opened 2022-01-05
fix the default value of predictor inputStd in android example

linkecoding:
In the Predictor class of the Android sample code, the default value of the inputStd variable is wrong, because the default value read from strings.xml is 0.485, 0.456, 0.406. Since the variable has been reassigned, this bug is not easy to find.


LDOUBLEV (approved):
LGTM

PR #5232, opened 2022-01-11
fix data to support offline augment

littletomatodonkey:
att.

* if the label is as follows.

```
["11.jpg", "12.jpg"]    test
```

then one of "11.jpg" and "12.jpg" will be chosen to train the model, which is often used for offline augmentation.


WenmuZhou:
应该在文档里里补一下 离线数据扩充的标注格式吧

littletomatodonkey:
> 应该在文档里里补一下 离线数据扩充的标注格式吧

done, thx

WenmuZhou (approved):

PR #5243, opened 2022-01-12
fix kie export

LDOUBLEV:
- [ ] add predict_kie.py


trucdtt:
@LDOUBLEV I try to export KIE model and run *predict_kie.py* file. However, there are some errors like: 
image

How can I fix it?

tools/infer/utility.py
```diff
@@ -145,6 +149,8 @@ def create_predictor(args, mode, logger):
         model_dir = args.rec_model_dir
     elif mode == 'table':
```
hthoai, line 150: try to comment this line

PR #5383, opened 2022-01-31
Added polish corpus and dict

krzynio:
Polish corpus from (processed and cleaned by myself):

http://zil.ipipan.waw.pl/DistrNKJP

The free subcorpus of the National Corpus of Polish consists of all its texts that are free from intelectual property constraints. These texts can be distributed without limitations.


JareelSkaj:
@krzynio comparing to the [Romanian corpus](https://github.com/PaddlePaddle/PaddleOCR/blob/0c937689585db4817ea7cbe55e735bc1a8196b44/ppocr/utils/corpus/romanian_corpus.txt) the one here seems off - it's got stuff like `o&amp;apos;reillym` or `ogólnymi"` or `3753` or `pn-r-75541` or `p-onadto`. At the very least it could use another cleanup sweep.

Also I would note that it's missing some basic words, e.g. `odloty` (departures - on the airport, pretty common thing to OCR), but I don't know how important it is.

skszymon20:
@krzynio Have You trained any text detector with a use of the corpus and the dictionary? If so, I would appreciate the contact with you or posting the trained model.

PR #5576, opened 2022-02-27
[android demo] Fix error when run same image

WenmuZhou:
This pr contains 
1. fix bug in https://github.com/PaddlePaddle/PaddleOCR/issues/5265 by upgrading PaddleLite to v2.10
2. removed load model button, models are now automatically loaded after related configuration changes. eg `CPU Thread Num` and `CPU Power Mode`.
3. the detection preprocessing is moved to the CPP side for execution
4. support 7 running modes
    * det+cls+rec
    * det+rec
    * cls+rec
    * det
    * rec
    * cls
 5. rm armv7 support

todo

- [x]  update v2.10 model

Sample image

 图片名称


LDOUBLEV (approved):
LGTM

Issue #5752, opened 2022-03-23
order_points_clockwise BUG


cqray1990:
请提供下述完整信息以便快速定位问题/Please provide the following information to quickly locate the problem

- 系统环境/System Environment：
- 版本号/Version：Paddle：  PaddleOCR： 问题相关组件/Related components：
- 运行指令/Command Code：
- 完整报错/Complete Error Message：

when the point is [[205,164],[230,166],[203,361],[178,362]]  the function results is wrong,i draw it in picture,it shows an cross lines,  as an results get_rotate_crop_image will get an wrong img,

cqray1990:
@MissPenguin @sdcb @ZeyuChen @bingooo

MissPenguin:
Thanks for your information. Bug fixed in this pr: https://github.com/PaddlePaddle/PaddleOCR/pull/5805

cqray1990:
@MissPenguin 
No,it did not slove this problem when the point is 
[[129,409],[348,204],[408,257],[181,462]], the result is not right

PR #5805, opened 2022-03-28
fix order_points_clockwise

MissPenguin:
fix cross lines problems mentioned in https://github.com/PaddlePaddle/PaddleOCR/issues/5752


WenmuZhou (approved):

cqray1990:
@MissPenguin @WenmuZhou 
No,it did not slove this problem when the point is
[[129,409],[348,204],[408,257],[181,462]], the result is not right

PR #5881, opened 2022-04-04
Romanian Corpus and Character set

the-ge:
The Romanian corpus file is a cleaned version of the official Romanian Scrabble word list (https://dexonline.ro/scrabble), licensed under GPL (https://dexonline.ro/licenta). In addition to the base form of the words, it contains the inflexions and the diacriticless form (diacritics are mostly not used online). Please let me know if the corpus should be simplified.
I'm not sure if that's anything else I should add. Here's the Wikipedia page about the Romanian language: https://en.wikipedia.org/wiki/Romanian_language


cristian-sima:
When will be this implemented?

Wyzix33:
Is this in the latest release?

the-ge:
Closed this as the CLA could not be signed and opened https://github.com/PaddlePaddle/PaddleOCR/pull/9456

PR #6361, opened 2022-05-20
Update README_en.md

mohamadmansourX:
Either: https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.5/deploy/lite/readme.md need to change the file name to README **_en** .md to match the pattern

Or this link should be to README.md without _en


Evezerest (approved):
LGTM

PR #6477, opened 2022-06-02
Fix copy_paste no texts augment.

MikoyChinese:
If you train data with CopyPaste and EastRandomCropData such as [ch_PP-OCRv3_det_cml.yml](https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.5/configs/det/ch_PP-OCRv3/ch_PP-OCRv3_det_cml.yml).

The original copy_paste only appends the new polys and new ignore_tags to data, but no new texts value to data.  In EastRandomCropData, the texts value is an important key in below https://github.com/PaddlePaddle/PaddleOCR/blob/473a811667adcc2706c8cfa6d04a66f9fbce02d7/ppocr/data/imaug/random_crop_data.py#L170-L179

If you tranform the train data with no new texts, you will get a wrong label. So append the texts to CopyPaste.

### DIFF
--- a/ppocr/data/imaug/copy_paste.py
+++ b/ppocr/data/imaug/copy_paste.py
@@ -35,10 +35,12 @@ class CopyPaste(object):
         point_num = data['polys'].shape[1]
         src_img = data['image']
         src_polys = data['polys'].tolist()
+        src_texts = data['texts']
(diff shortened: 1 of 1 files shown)


LDOUBLEV (approved):

LDOUBLEV:
LGTM

PR #6585, opened 2022-06-14
replace GPL licensed components

maxbachmann:
This PR replaces GPL licensed libraries for string comparisions with rapidfuzz which has the following advantages:
- MIT license compatible with the Apache License of PaddleOCR, which is not the case for the currently used GPL licensed libraries
- interface provides a normalized distance, which makes it possible to simplify the code here
- faster implementation than both python-Levenshtein and distance
- actively maintained, which is not the case both for python-Levenshtein and distance
- provides wheels for all common platforms, which simplifies the installation for users

### DIFF
--- a/ppocr/metrics/rec_metric.py
+++ b/ppocr/metrics/rec_metric.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import Levenshtein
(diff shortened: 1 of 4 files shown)


WenmuZhou (commented):
LGTM

GT-ZhangAcer:
Very good changes, and provides a whl package to avoid some user environment problems.

PR #6616, opened 2022-06-16
Fix TensorRT failed due to some TRT inputs dynamic shape info not set

Rm1n90:
After updating the paddleocr base code from version 2.4 to 2.5, TensorRT execution failed in the new version due to some TRT inputs dynamic shape info not set. By adding the following line the TensorRT should execute normally. Tested on 3 different machines :)


LDOUBLEV (approved):
LGTM

PR #6648, opened 2022-06-20
Fix SER and RE VQA docs command error

nbswords:
Fix the error path of End2End evaluation command of SER doc and OCR engine + SER + RE command


Evezerest:
pls cherry-pick this PR to dygraph branch

PR #6666, opened 2022-06-22
Updated WandbLogger to make uploading model checkpoints optional

manangoel99:
This PR adds a minor patch to the WandbLogger integration to make saving model checkpoints to the W&B dashboard optional. To save checkpoints to the dashboard, the users just need to update the yaml file like so

```
wandb:
    project: my_project
    entity: my_team
    log_checkpoint: true
```

Logging checkpoints is turned off by default.


manangoel99:
I was wondering if I could get some feedback on this PR whenever possible

manangoel99:
Hi @LDOUBLEV @WenmuZhou I was wondering if it would be possible to get a review on this PR

ppocr/utils/loggers/wandb_logger.py
```diff
@@ -40,12 +44,14 @@ def __init__(self,
...
             self.run.config.update(self.config)
+        
+        self.log_checkpoint = log_checkpoint
 
     @property
     def run(self):
         if self._run is None:
             if self.wandb.run is not None:
-                logger.info(
+                logging.info(
```
WenmuZhou, line 54: It is recommended to use the inner logger to print, which can be initialized with the following code
```python
from ppocr.utils.logging import get_logger
logger = get_logger()
```

TuanBC:
@manangoel99 This pull request is actually very useful for many use cases that I'm using. Could you review this pull request again based on the above comment?

manangoel99:
@WenmuZhou I've made the requested change

PR #7003, opened 2022-07-26
✨ Add Norwegian language dict

khellan:


andyjpaddle:
Thanks for your contribution! But the dict you provided is including in Latin dict https://github.com/PaddlePaddle/PaddleOCR/blob/dygraph/ppocr/utils/dict/latin_dict.txt

PR #7006, opened 2022-07-27
[KIE] add configurable ignore classes for KIEMetric

ChenNima:
There's a hard-coded ignores list for computing the F1 score while evaluating the model. If the class list of data is modified, the ignore list should be modified as well, otherwise, the F1 score could not reflect the actual performance of the model.

Instead of changing the hidden and hard-coded ignore class list, this change makes it configurable via the model config YAML file.


LDOUBLEV:
TODO: CP to 2.6 and dygraph

PR #7054, opened 2022-08-01
add use_xpu in config

helen88:
add use_xpu


helen88:
https://github.com/PaddlePaddle/PaddleOCR/blob/1696b36bdb4152138ed5cb08a357df8fe03dc067/tools/program.py#L569  already has this function.

PR #7120, opened 2022-08-05
Release/2.5

mohamadmansourX:
Fixed Loggers related functions in program.py
In addition the old script break when running and inference incase the config had wandb enabled since it was using save_model_dir parameter which is only created when is_train is true.


mohamadmansourX:
@manangoel99

WenmuZhou:
Hello, non-major bug class update, it is recommended to pr to dygraph minutes

luotao1:

PR #7277, opened 2022-08-21
Remove unmaintained attrdict dependency

hanacchi:
The `attrdict` library has been unmaintained since 2015 and is incompatible with newer Python versions (3.10+).
This pull request removes the dependency.


WenmuZhou:
你好，主分支已经没有使用attrdict了

PR #7807, opened 2022-10-03
Update predict_det.py

ArturoJ:
Avoid error if installed ( Tools )
Tested with ( tools==0.1.9 )


andyjpaddle (commented):
This change will cause inconsistencies between whl and local code when whl package is an old version. I suggest you uninstall `tools` or alias it.

PR #7810, opened 2022-10-03
Update predict_system.py

ArturoJ:
Avoid error if installed ( Tools )
Tested with ( tools==0.1.9 )


andyjpaddle (commented):
same with https://github.com/PaddlePaddle/PaddleOCR/pull/7807

PR #7872, opened 2022-10-11
Update README: Correct download recognition model URL

hs9021401:


andyjpaddle:
Thanks for your contribution! would you like cherry-pick this PR to dygraph branch?

PR #8102, opened 2022-10-25
Fix Visionlan model label encode

deisler134:
Visionlan model label always encode to lower letters


andyjpaddle:
thanks for your feedback, we have fixed this bug, see https://github.com/PaddlePaddle/PaddleOCR/pull/8128

PR #8217, opened 2022-11-06
Fix `CMAKE_CXX_FLAGS` optimize flag

Hirozy:
According to the gcc options, here should be an uppercase `O` instead of a lowercase `o`. And the lowercase `o` should not appear in `CMAKE_CXX_FLAGS`, which will cause the file to not be output correctly when compiled.

There are some details here.
[https://www.rapidtables.com/code/linux/gcc/gcc-o.html](https://www.rapidtables.com/code/linux/gcc/gcc-o.html)
[https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options)
[https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)


GreatV:
hi @Hirozy  Could you resubmit this PR to the main branch? We won't accept commits to release/2.6 unless absolutely necessary.

PR #8242, opened 2022-11-08
update predic_det.py for DB with "image_shape"

deisler134:
predict texts with default image_shape


luotao1:

PR #8268, opened 2022-11-10
remove order_points_clockwise, the prediction points should be clockw…

deisler134:
…ise and "order_points_clockwise" could cause some bugs when points have more than one max values


GreatV:
hi @deisler134 Could you resubmit this PR to the main branch? We won't accept commits to release/2.6 unless absolutely necessary.

PR #8414, opened 2022-11-22
changed opencv to opencv-headless

mani1709:
To fix error: 
    ImportError: libGL.so.1: cannot open shared object file: No such file or directory


victory-sokolov:
Hey!
Could we move forward with replacing `opencv-python` with` opencv-python-headless`?
Having problems installing dependencies in the Docker container that are required for `opencv-python`

PR #8499, opened 2022-11-30
removed xlrd as it does not support xlsx and moved to openpyxl

chiru-stefan:
Removed xlrd as it does not support xlsx and starts to get buggy with python 3.9; switched to openpyxl

### DIFF
--- a/PPOCRLabel/PPOCRLabel.py
+++ b/PPOCRLabel/PPOCRLabel.py
@@ -21,8 +21,8 @@ import os.path
 import platform
 import subprocess
 import sys
-import xlrd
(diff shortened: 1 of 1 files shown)


chiru-stefan:
"The 2.x series has explicitly removed support for anything other than xls files" - a workaround was to use xlrd==1.2.0
"At the low level the XML parsing in xlrd and and openpyxl is very similar, but the openpyxl code is far better tested and easier to
maintain." -> [xlrd vs openpyxl discussion](https://groups.google.com/g/python-excel/c/IRa8IWq_4zk)

PR #8593, opened 2022-12-09
Resolve DeprecationWarning for `np.int`

PythonTryHard:
This PR aims to resolve the `DeprecationWarning` raised due to usage of `np.int`.

`np.int` [has been deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), and migration is deemed safe, as per:
```
DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. 
To silence this warning, use `int` by itself. Doing this will not modify 
any behavior and is safe. When replacing `np.int`, you may wish to use e.g.
`np.int64` or `np.int32` to specify the precision. If you wish to review
your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: 
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

Furthermore, as this is only used in calculating box score, 32 bits are enough since we can safely assume the box's 4 coordinates are less than 2^32 - 1, as any coordinates that would require a 64-bit integer is highly likely the result of a bug rather than being the result of sane usage.

### DIFF
--- a/ppocr/postprocess/db_postprocess.py
+++ b/ppocr/postprocess/db_postprocess.py
@@ -185,10 +185,10 @@ class DBPostProcess(object):
         '''
         h, w = bitmap.shape[:2]
         box = _box.copy()
-        xmin = np.clip(np.floor(box[:, 0].min()).astype(np.int), 0, w - 1)
-        xmax = np.clip(np.ceil(box[:, 0].max()).astype(np.int), 0, w - 1)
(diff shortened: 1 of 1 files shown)


PythonTryHard:
Closed as PR has been addressed in 7e7630eb67ce77f7f65c6dab5bfbea2bebabb8d5.

PR #8605, opened 2022-12-13
Create train_test_split

harshnandwana:
This  code helps you converting output of PPOCRLabel to feed-able format that you can use for training and separate you single folder and single Label.txt file in train and val split


Evezerest:
Could you please write instructions in the readme of PPOCRLabel?

harshnandwana:
sure

rpsipl2021:
not working this train_test split

PR #8719, opened 2022-12-27
Fix: broken link for whl package documentation.

DhruvAwasthi:
This proposed change fixes the broken link under the section `2.4 Parameter Description` in the last line that says: Most of the parameters are consistent with the PaddleOCR whl package, see `whl package documentation`.


Evezerest:
Thanks for your contribution! please cherry-pick this commit to dygraph branch

PR #8720, opened 2022-12-27
Update parameter default type and value for image shapes

DhruvAwasthi:
The default image shape datatype for rec_image_shape and cls_image_shape is a list and the default value is also provided as a list argument. However, in the updated version, this is no longer the case. Now the default type is str, and the default value is also provided by a string.  Hence, updating these two parameter default datatype and value is required.


andyjpaddle (commented):
thanks for your contribution, we will fix it on dygraph branch firstly and then cherry-pick to release2.6 branch

DhruvAwasthi:
Okay. Should I create a PR In dygraph branch for the same?
Or do you want to do it yourself.

Thanks!

andyjpaddle:
> Okay. Should I create a PR In dygraph branch for the same? Or do you want to do it yourself.
> 
> Thanks!

I will fix it on both Chinese and English doc, so you can close this PR. 
Thanks a lot!

DhruvAwasthi:
> > Okay. Should I create a PR In dygraph branch for the same? Or do you want to do it yourself.
> > 
> > Thanks!
> 
> I will fix it on both Chinese and English doc, so you can close this PR. 
> Thanks a lot!

Sure. I'm closing this.
Thanks!

Issue #1024, opened 2020-10-27
No module named 'tools.infer' 


zzdts:

grasswolfs:
https://github.com/PaddlePaddle/PaddleOCR/blob/develop/doc/doc_ch/whl.md

littletomatodonkey:
你是按照哪个教程走，遇到了这个问题呢？好歹说下报错的步骤啊

zzdts:
ubuntu20.04  
conda 4.9.0  python3.7.7  
按照 [https://github.com/PaddlePaddle/PaddleOCR/blob/develop/doc/doc_en/installation_en.md](url) 这个教程，但是没有使用步骤1中的docker环境。  

这个错误很奇怪， 看了你们在这上面的讨论应该是解决了这个问题的，下载的源码中也有对应的commit id但是用了release分支和dev分支都还是一样。 
明明tools包就和引用的文件一个目录，为什么就不能导入？

littletomatodonkey:
你是运行了哪个命令报了上面的错误？

zzdts:
直接clone 包后 运行github上的sample时出现的。
现在发现直接使用python xx.py  运行程序没有问题，使用pdb xx.py   又出现这个问题:
`-> import studiowork
(Pdb)
ModuleNotFoundError: No module named 'tools.infer'
`

ATM006:
请问这个问题还有最新进展么！待续。。。

helindemeng:
自己写的代码中，存在名称为tools的包，导致paddleocr调用其本身的tools方法时，路径查询到我们自己写的tools了，所以把我们自己写的tools报名改一下即可。

milkliker:
@littletomatodonkey 
这个问题主要还是paddle代码的tools没有绝对引用带来的问题。
建议把paddle代码里
`import tools...`
都改为
`import paddleocr.tools...`
避免与开发者自己的tools冲突

lvguangtao:
@milkliker 我应该去哪个脚本里修改呢

sallenhandong:
> 自己写的代码中，存在名称为tools的包，导致paddleocr调用其本身的tools方法时，路径查询到我们自己写的tools了，所以把我们自己写的tools包名改一下即可。

yes,it works

jimmyflycv:
> 

tried but still the error exists.

stevenong99:
[...]

PR #8968, opened 2023-02-03
Fix "No module named 'tools.infer'" error #1024

HuayuanSong:
Further specified that tools refers to the tools from PaddleOCR to prevent conflicting packaging names, hence tools being a pretty generic name.


GreatV:
@HuayuanSong Could you resubmit this PR to the main branch? We won't accept commits to release/2.6 unless absolutely necessary.

PR #8744, opened 2022-12-30
Fix "IndexError: list index out of range" exception when make_border_map

TuanBC:
Reference:
https://github.com/WenmuZhou/PytorchOCR/commit/85acb9e4571b9af4de59774ffd6f2a90b9231828

PR #9353, opened 2023-03-08
Fix exception when make_border_map

TuanBC:
Fix the exception "IndexError: list index out of range"  that can happen during training detection model

Reference:
https://github.com/WenmuZhou/PytorchOCR/commit/85acb9e4571b9af4de59774ffd6f2a90b9231828
https://github.com/PaddlePaddle/PaddleOCR/pull/8744


LDOUBLEV (approved):
LGTM

PR #7933, opened 2022-10-14
Add Vietnamese dictionary and corpus files

rama298:
https://github.com/PaddlePaddle/PaddleOCR/issues/1048 

I have add the files which you asked to tarin the paddleocr for Vietnamese language

Issue #1048, opened 2020-10-28
Multilingual OCR development Plan


grasswolfs:
1、Chinese（done）
2、English（done）
3、French（done）
4、German（done）
5、Korean（done）
6、Japanese（done）
7、Russian（ongoing）
8、Italian（ongoing）

saheya:
Traditional Mongolian

omar16100:
I would love to work on "Bangla"

levanpon98:
I very happy if you do that with Vietnamese

HusseinYoussef:
How about Arabic? That would be great.

Hieung28:
I've find out that PADDLE OCR algorithm cannot recognize some special characters (such as comma, semicolon, or dot...) when the language is english. Is there any possible way that i can fix this problem

GmGniap:
I would like to contribute to add the Burmese language. Is it only needed to submit two text files - _dict_ & _corpus_? How further process do we need to provide?

shahidul56:
Adding "Bangla" will be  grate for the people in south Asia

giranntu:
Adding "Traditional Chinese (zh-TW)" would be great support.

Ru-Van:
Do you have preTrained Russian recognition model?

SasiAravind:
Hi  adding " Tamil" language will be very grateful.

[Tamil_dict.txt](https://github.com/JaidedAI/EasyOCR/blob/master/easyocr/character/ta_char.txt)
[Tamil_corpus.txt](https://github.com/JaidedAI/EasyOCR/blob/master/easyocr/dict/ta.txt)

Need more help plz refer this issue:
https://github.com/JaidedAI/EasyOCR/issues/39

fcakyon:
Turkish

krzynio:
I can help with polish language.

xmy0916:
@GmGniap Hello, Can you provide the corpus file of Burmese Language？

xmy0916:
@shahidul56 Hello, Can you provide the corpus file of Bangla Languag？

azmat21:
All models updated in 2021.1.21 cannot be downloaded with following Error：
`{
code: "NoSuchKey",
message: "The specified key does not exist.",
requestId: "aa1bfeff-f572-40aa-8935-6129b1533ed1"
}`

grasswolfs:
> All models updated in 2021.1.21 cannot be downloaded with following Error：
[...]

PR #9456, opened 2023-03-15
Romanian Corpus and Character set

the-ge:
The Romanian corpus file is a cleaned version of the official Romanian Scrabble word list (https://dexonline.ro/scrabble), licensed under GPL (https://dexonline.ro/licenta). In addition to the base form of the words, it contains the inflexions and the diacriticless form (diacritics are mostly not used online). Please let me know if the corpus should be simplified.
I'm not sure if that's anything else I should add. Here's the Wikipedia page about the Romanian language: https://en.wikipedia.org/wiki/Romanian_language.

This is the same as https://github.com/PaddlePaddle/PaddleOCR/pull/5881, which I closed because the pull request was done with a different email address, which in turn didn't let me sign the CLA.


the-ge:
Please provide some feedback as to what more needs to be done to merge the Romanian corpus and character set. I see that the Vietnamese PR (#7933) is in the limbo as well. Clearly, the steps outlined in  Multilingual OCR Development Plan (#1048) are not enough.

PR #9791, opened 2023-04-21
Update VQA to use the updated LayoutLM syntax from PaddleNLP

sijunhe:
Update vqa to use the updated LayoutLM syntax from PaddleNLP


tink2123 (approved):
LGTM

PR #9980, opened 2023-05-18
fixbug: change DetOp.postprocess's result from string(ndarray) to string(list)

diana-pwf:
image
In web_service_det.py, DetOp.postprocess function, I find dt_boxes is a ndarray and it is converted to string, but the string representation 'str(a)' of a numpy array "a" is not sufficient to exactly restore the original array, particularly for arrays with many elements. This is because the string representation may truncate large arrays for brevity. So I change it to a list converted to string.

### DIFF
--- a/deploy/pdserving/web_service_det.py
+++ b/deploy/pdserving/web_service_det.py
@@ -61,8 +61,8 @@ class DetOp(Op):
         ]
         dt_boxes_list = self.post_func(det_out, [ratio_list])
         dt_boxes = self.filter_func(dt_boxes_list[0], [self.ori_h, self.ori_w])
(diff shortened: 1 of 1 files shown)


xiaoliou008:
I completely agree with this Pull Request. I've encountered a similar issue before, where the conversion of the ndarray to a string representation using 'str(a)' resulted in potential data loss due to truncation, especially for arrays with a large number of elements. Converting the ndarray to a list before converting it to a string is a better approach as it preserves the entire array without truncation. This change will ensure the accuracy and integrity of the data. Thank you for addressing this issue and proposing the solution!

youkaichao:
This PR saved my day! I use paddleocr for a long text, and the `str(dt_boxes)` caused all the trouble.

PR #10017, opened 2023-05-23
bug fix : gen rec labels split line

dhirajms:
removing space from ground truth and splitting at first "," is actually removing after the comma.
example -
xyz.png,dhi,raj

here filename is "xyz.png"
and text is "dhi,raj"

but with current code it's only taking "dhi"
also why removing space from ground truth?


DhirajMS:
@XInitialize

PR #10121, opened 2023-06-08
【benchmark】fix new profiler 

mmglove:
【benchmark】fix new profiler
    # profile : https://www.paddlepaddle.org.cn/documentation/docs/zh/guides/performance_improving/profiling_model.html#chakanxingnengshujudetongjibiaodan

cmd:
python tools/train.py ...
--profiler_options="batch_range=[10,20];state=GPU;tracer_option=Default;profile_path=model.profile;timer_only=True"

python tools/train.py ...
--profiler_options="batch_range=[10,20];state=GPU;tracer_option=Default;profile_path=model.profile;timer_only=False"

### DIFF
--- a/ppocr/utils/profiler.py
+++ b/ppocr/utils/profiler.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import sys
-import paddle
+import paddle.profiler as profiler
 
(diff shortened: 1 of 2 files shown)


tink2123 (approved):
LGTM

PR #10203, opened 2023-06-19
fix the example code at visualization 

ngoanpv:
In the current code, it has the error `TypeError: '<' not supported between instances of 'tuple' and 'float'
` because it's not looped to the each line value in the `result`


shiyutang:
There is a lot of conflict, did you propose your changes to the right branch

shiyutang:
close due to become stale.

PR #10216, opened 2023-06-20
Don't break overall processing on a bad image

UserUnknownFactor:
When processing multiple images PaddleOCR currently breaks if even a single image is malformed. This is to prevent that.

### DIFF
--- a/tools/infer/predict_system.py
+++ b/tools/infer/predict_system.py
@@ -65,15 +65,25 @@ class TextSystem(object):
         self.crop_image_res_index += bbox_num
 
     def __call__(self, img, cls=True):
-        time_dict = {'det': 0, 'rec': 0, 'csl': 0, 'all': 0}
+        time_dict = {'det': 0, 'rec': 0, 'cls': 0, 'all': 0}
+
+        if img is None:
+            logger.debug("no valid image provided")
+            return None, None, time_dict
+
         start = time.time()
         ori_im = img.copy()
         dt_boxes, elapse = self.text_detector(img)
         time_dict['det'] = elapse
-        logger.debug("dt_boxes num : {}, elapse : {}".format(
-            len(dt_boxes), elapse))
+
         if dt_boxes is None:
-            return None, None
+            logger.debug("no dt_boxes found, elapsed : {}".format(elapse))
+            end = time.time()
+            time_dict['all'] = end - start
+            return None, None, time_dict
+        else:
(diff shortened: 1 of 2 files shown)


tools/infer/predict_system.py
```diff
@@ -65,15 +65,21 @@ def draw_crop_rec_res(self, output_dir, img_crop_list, rec_res):
         self.crop_image_res_index += bbox_num
 
     def __call__(self, img, cls=True):
+        if img is None:
```
shiyutang, line 68: The assertation here prevent img from being None:https://github.com/PaddlePaddle/PaddleOCR/blob/fac03876f39bc67acc8eef6d7facb9a2206eeecd/paddleocr.py#L522

shiyutang (changes requested):
There are unnecessary code need to be removed.

tools/infer/predict_system.py
UserUnknownFactor: Fixed.

UserUnknownFactor: @shiyutang: this is actually incorrect as, let's say `img` is `str` so it's not `None` and assertion is passed there, but then we load it in `check_img` and if it's broken then the method returns `None`. So `img` is `None` in that case.

shiyutang: You are right, especially when img is provided in list, it is not checked.

tools/infer/predict_system.py
```diff
@@ -65,15 +65,26 @@ def draw_crop_rec_res(self, output_dir, img_crop_list, rec_res):
         self.crop_image_res_index += bbox_num
 
     def __call__(self, img, cls=True):
-        time_dict = {'det': 0, 'rec': 0, 'csl': 0, 'all': 0}
+        time_dict = {'det': 0, 'rec': 0, 'cls': 0, 'all': 0}
+
+        if img is None:
+            logger.debug("no valid image provided")
+            time_dict['all'] = 0
```
shiyutang, line 72: This line is redundant since time_dict only records the time elapse of this image, which is 0 as how it is initialized.

shiyutang (changes requested):

shiyutang (approved):
LGTM

shiyutang:
We are now holding a contribution to Paddleseg and Paddle OCR activity, you are welcome to join:[#10223](https://github.com/PaddlePaddle/PaddleOCR/issues/10223)

PR #10217, opened 2023-06-20
Add preprocessing common to OCR tasks

UserUnknownFactor:
Common OCR tasks often include filling transparent areas with actual color, inverting image or its binarization. This commits adds those as optional parameters for `ocr` function.


LDOUBLEV (approved):

paddleocr.py
```diff
@@ -539,6 +542,11 @@ def ocr(self, img, det=True, rec=True, cls=True):
         if det and rec:
             ocr_res = []
             for idx, img in enumerate(imgs):
+                img = alpha_to_color(img, alpha_color)
+                if inv:
+                    img = cv2.bitwise_not(img)
+                if bin:
+                    img = binarize_img(img)
```
shiyutang, line 549: Why is the preprocess only exist in the both det and rec scenario?

shiyutang (commented):
The solution is solid but I feel it is a bit redundant to expose all the image processing params in PaddleOCR.ocr，do we have a better solution to

paddleocr.py
UserUnknownFactor: Because it's can be helpful in both. Are there some other places it'd be useful? 
Please elaborate on your point.

shiyutang: 1. Could inv/bin be useful in detection-only and rec-only scenarios?
2. How about exposing all the image processing params in the init_args like the postprocessing params? https://github.com/PaddlePaddle/PaddleOCR/blob/fac03876f39bc67acc8eef6d7facb9a2206eeecd/ppstructure/utility.py#L21

UserUnknownFactor:
@shiyutang How about this?

paddleocr.py
UserUnknownFactor: I understand now. See the next commit for the changes.

UserUnknownFactor:
> Args is passed into PaddleOCR in L662, therefore the preprocess args is already in the engine and can be accessed

But what if we want to use those options through the API and not from the console application parameters? Won't this make things difficult because we'll need to reconfigure engine parameters then?

shiyutang:
In the above way, if we need to use image preprocess options through API, we can directly pass the params into PaddleOCR.

```
PaddleOCR(bin=True,..)
```

shiyutang (commented):
On top of code change, we may also need to update the docs. https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.6/doc/doc_ch/inference_args.md
https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.6/doc/doc_en/inference_args.md

UserUnknownFactor:
>PaddleOCR(bin=True,..)

It will prevent us from changing the settings on per-file basis, and it is needed sometimes.  
So I think the current implementation is the best compromise.

>[doc_en/inference_args.md](https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.6/doc/doc_en/inference_args.md)

This doesn't exist. Do you want me to create it? I don't know Chinese, though...

shiyutang:
OK~ you can add here:https://github.com/PaddlePaddle/PaddleOCR/blob/dygraph/doc/doc_en/inference_args_en.md

UserUnknownFactor:
@shiyutang: Done.

UserUnknownFactor:
@shiyutang: can you please cherrypick GH-10217 and GH-10216 to PaddlePaddle:dygraph and  PaddlePaddle:release/2.7 if possible?

shiyutang:
Do you have any problem doing that? I can review for you~

UserUnknownFactor:
@shiyutang: No problem: GH-10654, GH-10655. 
By the way can you please explain why is 2.7 version not based off 2.6? Is there some different approach to versioning?

shiyutang:
2.7 is the snapshot of the dygraph branch， because we added lots of bugfix and new features on dygraph, it is easy to checkout a new branch on it.

PR #10249, opened 2023-06-27
[MLU] add mlu device for infer

PeiyuLau:
add mlu device for infer

### DIFF
--- a/tools/infer/utility.py
+++ b/tools/infer/utility.py
@@ -30,8 +30,10 @@ from ppocr.utils.logging import get_logger
 def str2bool(v):
     return v.lower() in ("true", "yes", "t", "y", "1")
 
+
 def str2int_tuple(v):
     return tuple([int(i.strip()) for i in v.split(",")])
+
 
 def init_args():
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


tools/infer/utility.py
onecatcn, line 1: There is some discrepency between the codes for npu and mlu, please justify the difference.

tools/infer/utility.py
```diff
@@ -245,8 +249,9 @@ def create_predictor(args, mode, logger):
                     logger.info(E)
                     logger.info("Please keep your paddlepaddle-gpu >= 2.3.0!")
 
-        elif args.use_npu:
-            config.enable_custom_device("npu")
+        elif args.use_custom_device:
+            dev_type = paddle.device.get_all_custom_device_type()[0]
+            config.enable_custom_device(dev_type)
```
shiyutang, line 254: The custom device should not be able to replace npu.

It seems you should add mlu device on top of the original code with another elif clause.

shiyutang (changes requested):

tools/infer/utility.py
```diff
@@ -37,6 +37,11 @@ def init_args():
     parser.add_argument("--use_gpu", type=str2bool, default=True)
     parser.add_argument("--use_xpu", type=str2bool, default=False)
     parser.add_argument("--use_npu", type=str2bool, default=False)
+    parser.add_argument(
+        "--use_custom_device",
```
shiyutang, line 41: If you want to add mlu, I suppose it is better if this should is use_mlu.

tools/infer/utility.py
shiyutang: config.enable_custom_device("mlu")

shiyutang (changes requested):
any custom device is vague and should be replaced with mlu

shiyutang (approved):
LGTM

PR #10251, opened 2023-06-27
fix undefined save_model_dir when using wandb

itasli:


shiyutang (changes requested):
Hello, Thanks for your contribution but I do not see the point of changing the variable name, and it is not correct because save_dir is used in the following code. image

itasli:
> Hello, Thanks for your contribution but I do not see the point of changing the variable name, and it is not correct because save_dir is used in the following code. image

Hey thanks for you reply, actually I wanted to keep the change consistent with the rest of the code and use `save_model_dir`, but indeed we can also keep `save_dir` and modify only line 688.

shiyutang (approved):
LGTM

shiyutang:
Hi, We are now having the contribute to the OCR and Seg activity, you are welcome to join: https://github.com/PaddlePaddle/PaddleOCR/issues/10223

PR #10323, opened 2023-07-07
fix backbone weights load

TimeYWL:


ppocr/utils/save_load.py
```diff
@@ -153,26 +153,27 @@ def load_pretrained_params(model, path):
...
     new_state_dict = {}
     is_float16 = False
 
-    for k1 in params.keys():
-
-        if k1 not in state_dict.keys():
-            logger.warning("The pretrained params {} not in model".format(k1))
+    map_dict = MapBackboneWeights(params, state_dict)
+    
+    for k1 in map_dict.keys():
```
shiyutang, line 162: The thing with map_dict is that it only load backbone params. Therefore it will only have backbone params in new_state_dict as well, which is not what we expect in loading a model.

I suggest adding params not in the backbone in the map_dict as well, but only map to themselves.

ppocr/utils/save_load.py
```diff
@@ -153,26 +153,27 @@ def load_pretrained_params(model, path):
...
+            if params[k1].dtype != state_dict[k2].dtype:
+                params[k1] = params[k1].astype(state_dict[k2].dtype)
+            if list(state_dict[k2].shape) == list(params[k1].shape):
+                new_state_dict[k2] = params[k1]
             else:
                 logger.warning(
                     "The shape of model params {} {} not matched with loaded params {} {} !".
-                    format(k1, state_dict[k1].shape, k1, params[k1].shape))
-
+                    format(k1, state_dict[k2].shape, k1, params[k1].shape))
```
shiyutang, line 176: I suppose this should be 
```
format(k2, state_dict[k2].shape, k1, params[k1].shape))
```
since only k2 is 100% sure in  the state_dict.

shiyutang (changes requested):
There are a few bugs.

andyjpaddle:
@TimeYWL I suggest that you rename the keys name of weights offline, then load it.

shiyutang:
close due to become stale.

PR #10352, opened 2023-07-11
feature : add normalized Character Error Rate

itasli:
Based on Levenshtein distance, CER represents the percentage of characters in the reference text that was incorrectly predicted in the OCR output. The lower the CER value the better the performance (0 being a perfect score).

### DIFF
--- a/ppocr/metrics/rec_metric.py
+++ b/ppocr/metrics/rec_metric.py
@@ -41,23 +41,35 @@ class RecMetric(object):
         correct_num = 0
         all_num = 0
         norm_edit_dis = 0.0
-        for (pred, pred_conf), (target, _) in zip(preds, labels):
+        edit_dis = 0
+        
+        for (pred, _), (target, _) in zip(preds, labels):
+            
             if self.ignore_space:
                 pred = pred.replace(" ", "")
                 target = target.replace(" ", "")
             if self.is_filter:
                 pred = self._normalize_text(pred)
                 target = self._normalize_text(target)
+
             norm_edit_dis += Levenshtein.normalized_distance(pred, target)
+            edit_dis += Levenshtein.distance(pred, target)
+            
             if pred == target:
                 correct_num += 1
+            
             all_num += 1
+
+
         self.correct_num += correct_num
         self.all_num += all_num
         self.norm_edit_dis += norm_edit_dis
+        self.edit_dis += edit_dis
+
         return {
             'acc': correct_num / (all_num + self.eps),
-            'norm_edit_dis': 1 - norm_edit_dis / (all_num + self.eps)
+            'norm_edit_dis': 1 - norm_edit_dis / (all_num + self.eps),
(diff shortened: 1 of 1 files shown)


ppocr/metrics/rec_metric.py
```diff
@@ -41,41 +41,56 @@ def __call__(self, pred_label, *args, **kwargs):
...
+            
             if self.ignore_space:
                 pred = pred.replace(" ", "")
                 target = target.replace(" ", "")
             if self.is_filter:
                 pred = self._normalize_text(pred)
                 target = self._normalize_text(target)
+
             norm_edit_dis += Levenshtein.normalized_distance(pred, target)
+            edit_dis += Levenshtein.distance(pred, target)
```
shiyutang, line 56: As it is shown in the doc normalize_distance = distance / max, therefore they are  interchangeable. So why do we need to add distance on top of the normalize_distance we already have?

refer: https://maxbachmann.github.io/RapidFuzz/Usage/distance/Levenshtein.html#rapidfuzz.distance.Levenshtein.normalized_distance

shiyutang (commented):
Please show your reasons of adding this metric.

ppocr/metrics/rec_metric.py
itasli: Hey @shiyutang, first of all here is [link](https://towardsdatascience.com/evaluating-ocr-output-quality-with-character-error-rate-cer-and-word-error-rate-wer-853175297510#5aec) to an article about the metric I'd like to implement.

As explained there **CER** calculation is based on the concept of **Levenshtein** distance, and is defined as :

CER<sub>normalized</sub> = S+D+I / S+D+I+C

* S + D + I, being **Levenshtein** distance (not normalized) referred as `edit_dis`in the code
* C, being number of correct characters, referred as `correct_num` in the code

where:
   * S = Number of Substitutions
   * D = Number of Deletions
   * I = Number of Insertions

Simplifying a bit would give us either :
* CER<sub>normalized</sub> = `Levenshtein.distance(pred, target)` / `Levenshtein.distance(pred, target)` + C
or
* CER<sub>normalized</sub> = `Levenshtein.distance(pred, target)` / `len(target)` + I

I totally agree that computing Levenshtein distance twice isn't a great solution, and given the way it is implemented in RapidFuzz, i can rework my PR to compute either non normalized Levenshtein distance **OR** normalized Levenshtein distance and use [`_levenshtein_maximum`](https://github.com/maxbachmann/RapidFuzz/blob/main/src/rapidfuzz/distance/Levenshtein_py.py#L14) method to compute the `max_distance`.

shiyutang:
We are now holding a contribution to Paddleseg and Paddle OCR activity, you are welcome to join:https://github.com/PaddlePaddle/PaddleOCR/issues/10223

itasli:
After a few calculations and iterations, I realize that the character error rate metric I'm trying to calculate is equal to `Levenshtein.normalized_distance(pred, target)`, and that basically `norm_edit_dist` is equivalent to `normalized_silimarity()` method...

PR #10418, opened 2023-07-17
reverse output for rtl languages has bug.

masoudMZB:
paddle main code to reverse label for RTL languages has bug, try to use [python-bidi](https://pypi.org/project/python-bidi/). this works fine

### DIFF
--- a/ppocr/postprocess/rec_postprocess.py
+++ b/ppocr/postprocess/rec_postprocess.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from bidi.algorithm import get_display
 import numpy as np
 import paddle
 from paddle.nn import functional as F
@@ -49,20 +50,7 @@ class BaseRecLabelDecode(object):
         self.character = dict_character
 
     def pred_reverse(self, pred):
-        pred_re = []
-        c_current = ''
-        for c in pred:
-            if not bool(re.search('[a-zA-Z0-9 :*./%+-]', c)):
-                if c_current != '':
-                    pred_re.append(c_current)
-                pred_re.append(c)
-                c_current = ''
-            else:
-                c_current += c
-        if c_current != '':
-            pred_re.append(c_current)
-
-        return ''.join(pred_re[::-1])
+        return get_display(pred)
 
     def add_special_char(self, dict_character):
         return dict_character


shiyutang (commented):
Thank you for your contribution, we will look into it ASAP. Although can you kindly point out what is the bug of current code?

masoudMZB:
Thanks for you attention @shiyutang @ToddBear 
Sure, I'll show some errors of current code as soon as I get home.

ToddBear:
186a4f69a6a845d2c7706349d605ee39

I tested an example, but found not difference between the outputs of two implements. Can you provide some special cases ?

andyjpaddle:
> Thanks for you attention @shiyutang @ToddBear Sure, I'll show some errors of current code as soon as I get home.

@masoudMZB Have you test some bad case for current code? if there is no difference, we will close this PR

Husain98a:
I'm not sure about the bug @masoudMZB is referring to, but I've also noticed a problem with this function.
Paddle Post processing function[ pred_reverse()](https://github.com/PaddlePaddle/PaddleOCR/blob/1ab1e3364a31fc5c9b0663a5c0665a81f45c231a/ppocr/postprocess/rec_postprocess.py#L51) is supposed to reverse charecters of RTL languages (e.g. Arabic) including arabic numerals. The problem is that numerals in Arabic are written from left to right, while letters are written from right to left, so this function will produce reversed arabic numbers which is wrong.  
 [python-bidi](https://pypi.org/project/python-bidi/)  is a better approuch
example:
image

masoudMZB:
@andyjiang1116 sure check these examples
Text recognition models predict output from the most left character in the image. while RTL languages ordering is reverse for characters but LTR for numbers ( a little weird ). when numerical characters, alphabetical characters, punctuations come together in an image, paddle code will not produce the proper ordering.


Paddle default code output is as bellow ( as you see ordering is not correct ) ( to understand the ordering you need to be familiar with Persian/Arabic Language ) :

> Input: ،۵۱۳۸۵-۳۴۱۵
 Output: ۵۱۴۳-۵۸۳۱۵،
 ======
Input: ۹۳/۱۰/۲۹
Output: ۹۲/۰۱/۳۹
 ======
Input: ۳۵۸/د
Output: د/۸۵۳

while when using `from bidi.algorithm import get_display` the output and ordering are correct : 
> Input: ،۵۱۳۸۵-۳۴۱۵
  Output: ،۵۱۳۸۵-۳۴۱۵
 ======
Input: ۹۳/۱۰/۲۹
Output: ۹۳/۱۰/۲۹
 ======
Input: ۳۵۸/د
Output: د/۳۵۸

there are much more examples, you can test them by yourself

chenjjcccc:
> ۳۵۸/د
yep,I reproduced the same problem
reverse

PR #10421, opened 2023-07-17
fix "No module named xxx" error after built target executable file by Pyinstaller and run.

kerneltravel:
fix "No module named xxx" error after built target executable file by Pyinstaller and run. the related issue ID is #10398, #10340, #9951, #9723 #9694. according to the suggestion from issue #9408.

Issue #10398, opened 2023-07-14
ModuleNotFoundError: No module named 'tools'


xuri422:
My code run in JetBrain is very well, There are no errors.

But when I use "pyinstaller -F main.py"  build code to exe file(no error), and then run in command line. there are show me some errors:

PyInstaller\loader\pyimod02_importers.py:385: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
Traceback (most recent call last):
  File "main.py", line 5, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 385, in exec_module
  File "paddleocr\__init__.py", line 14, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 385, in exec_module
  File "paddleocr\paddleocr.py", line 33, in <module>
  File "importlib\__init__.py", line 126, in import_module
ModuleNotFoundError: No module named 'tools'
[11752] Failed to execute script 'main' due to unhandled exception!

I have try it in two PC, show the some errors. So, How can I resolve it?

xuri422:
pip list show versions:
paddleocr                 2.6.1.3
paddlepaddle           2.4.2
pandas                     2.0.1

UserWangZz:
https://github.com/PaddlePaddle/PaddleOCR/discussions/11490
you can reference this link to packpage the code. the paddle official isn't support to package the paddleocr.

Issue #10340, opened 2023-07-10
Pyinstaller - no module named "ppocr"


olqi:
请提供下述完整信息以便快速定位问题/Please provide the following information to quickly locate the problem

- 系统环境/System Environment：windows 10
- 版本号/Versin: Paddle：2.5.0  PaddleOCR：2.6.1.3 问题相关组件/Related components：vscode, python 3.9.2

Hey, when building an exe with Pyinstaller from a python file that uses PaddleOCR I get a "No module named 'ppocr' error" (attached a screenshot). How do I fix it?

The code I'm using is just the most barebones one to find out how to create an exe file with PaddleOCR:

`import paddleocr

ocr = paddleocr.PaddleOCR()

path = input('input path: ')

result = ocr.ocr(path)

print(result)`

Thank you! 
![image](https://github.com/PaddlePaddle/PaddleOCR/assets/139183372/9941d368-54a6-4d58-9c22-359179d0c03b)

Edmond-Lee-Zse-Wong:
I got another but similar error when I use "from paddleocr import PaddleOCR",  my error is "FileNotFoundError"
![Uploading 屏幕截图 2023-12-23 220522.png…]()

Edmond-Lee-Zse-Wong:
I got another but similar error when I use "from paddleocr import PaddleOCR", my error is "FileNotFoundError"
![屏幕截图 2023-12-23 220522](https://github.com/PaddlePaddle/PaddleOCR/assets/56387202/61d7011c-c19d-4cb0-a1af-3a3dea5948b7)

kerneltravel:
@Edmond-Lee-Zse-Wong  according to [this method](https://github.com/PaddlePaddle/PaddleOCR/pull/10502), you'd try to append parameter `--collect-all  paddleocr` when package with module paddleocr by pyinstaller .

Heryk13:
Here i made a tutorial for making a exe with pyinstaller

https://github.com/PaddlePaddle/PaddleOCR/discussions/11490#discussion-6072471

UserWangZz:
该issue长时间未更新，暂将此issue关闭，如有需要可重新开启。

Issue #9951, opened 2023-05-15
Pyinstaller exe problem


hsk20011:
File "paddleocr\__init__.py", line 14, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "paddleocr\paddleocr.py", line 33, in <module>
  File "importlib\__init__.py", line 126, in import_module
ModuleNotFoundError: No module named 'tools'

olqi:
i know its kinda late but if you'r still lloking for a solution, it's here - https://github.com/PaddlePaddle/PaddleOCR/issues/1024#issuecomment-1105946934

Heryk13:
Try this maybe it will work, if dont coment there maybe i can help you

https://github.com/PaddlePaddle/PaddleOCR/discussions/11490#discussion-6072471

UserWangZz:
This issue has not been updated for a long time. This issue is temporarily closed and can be reopened if necessary.

PR #10502, opened 2023-07-30
remove un-used imported modules, so to avoid PyInstaller packaged binary runtime error

kerneltravel:
by current PR, remove un-used imported modules in code, so to avoid PyInstaller packaged binary 's py-script import them, which in fact is not included by pyinstaller when package. 

if not merge current PR , then PyInstaller should add parameter `--hidden-import imghdr --hidden-import scipy.io` and must add parameter `--collect-all  paddleocr` , but the previous parameter is not suggested to hidden-import un-used modules for package release. 

Previously there is a [commit](https://github.com/PaddlePaddle/PaddleOCR/commit/32fe3c9b07f125b0c2e47d6a47776d128ee6847b) has `Improve package import compatibility` of paddleocr.py file. but the pyinstaller packaged execution file runtime not found module issue still exists.

I had tried another PR https://github.com/PaddlePaddle/PaddleOCR/pull/10421 ,but that is not the root cause.

history:
1. the related issues ID #1024 #9355 #10398, #10340, #9951, #9723 #9694, #9944, #9408 , #7635, #4390
2. how to reproduce the above issue: https://github.com/kerneltravel/PaddleOCR_PR10421_reproduce/blob/1ed3c8211ca07f9779c10e07dfbfcf81466ddfd5/README.md

NOTE: those people who want to use pyinstaller  to package python-application with paddleocr may referer to [this](https://github.com/kerneltravel/PaddleOCR_PR10421_reproduce/commit/7ccfed344e88e371ee744a335e7dd4b3fff0ffca) to  append parameter `--collect-all paddleocr `  to pyinstaller . `--collect-all MODULE_NAMES` means `Collect all submodules, data files, and binaries from the specified package or module`.

### DIFF
--- a/ppocr/data/imaug/ct_process.py
+++ b/ppocr/data/imaug/ct_process.py
@@ -20,7 +20,6 @@ import paddle
 
 import numpy as np
 import Polygon as plg
-import scipy.io as scio
 
 from PIL import Image
 import paddle.vision.transforms as transforms
--- a/ppocr/utils/utility.py
+++ b/ppocr/utils/utility.py
@@ -14,7 +14,6 @@
 
 import logging
 import os
-import imghdr
 import cv2
 import random
 import numpy as np


shiyutang (approved):
LGTM

shiyutang:
Thank you very much for your contribution, please cherry-pick this commit to the dygraph branch as well. :)

kerneltravel:
@shiyutang 要求怎样做？

> Thank you very much for your contribution, please cherry-pick this commit to the dygraph branch as well. :)

Issue #7823, opened 2022-10-05
Potential memory leak in PaddleOCR?


nikos-livathinos:
I have noticed some weird memory usage when evaluating the performance of PaddleOCR:

- When PaddleOCR processes new images of a sequence, there is a **constant increase in memory usage** of the process.
- The time profile of the memory usage has sudden steps to higher memory levels.
- If the system sets an upper bound in the memory consumption, the paddle process is eventually killed. It won't give allocated memory back, ever.
- This behaviour can be observed both with the C++ and the Python interface.

Eventually **it is impossible to keep running a PaddleOCR process as a service because the system runs out of memory and the process is killed.**

Can you provide insights on this memory usage pattern? Do you have any remedy? 

The following sections describe the tests in detail.

### Setup for the C++ tests

- Test hardware:
  - OS: Ubuntu 20.04
  - CPU: Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, 16 threads, AVX2
  - RAM: 128GB
- C++ compiler: `gcc v9.4.0`

- PaddleOCR: Source code from `v2.6.0`
- Paddle library:

   ```
   GIT COMMIT ID: 4596b9a22540fb0ea5d369c3c804544de61d03d0
   WITH_MKL: ON
   WITH_MKLDNN: ON
   WITH_GPU: OFF
   WITH_ROCM: OFF
   WITH_ASCEND_CL: OFF
   WITH_ASCEND_CXX11: OFF
   WITH_IPU: OFF
   CXX compiler version: 8.2.0
   ```
- Model files:
  - Detector: `en_PP-OCRv3_det_infer.tar`
  - Classifier: `ch_ppocr_mobile_v2.0_cls_infer.tar`
  - Recognizer: `en_PP-OCRv3_rec_infer.tar`

- OpenCV has been compiled from the source code tagged at `v4.6.0` with the parameters:

    ```
    cmake ../opencv \
        -DCMAKE_INSTALL_PREFIX=${install_dir} \
        -DCMAKE_BUILD_TYPE=Release \
        -DBUILD_SHARED_LIBS=OFF \
        -DWITH_IPP=OFF \
        -DBUILD_IPP_IW=OFF \
        -DWITH_LAPACK=OFF \
        -DWITH_EIGEN=OFF \
        -DCMAKE_INSTALL_LIBDIR=lib64 \
        -DWITH_ZLIB=ON \
        -DBUILD_ZLIB=ON \
[...]

Issue #7155, opened 2022-08-09
 I'm looking for a way to limit memory usage in CPU mode


Luyin:
请提供下述完整信息以便快速定位问题/Please provide the following information to quickly locate the problem

- 系统环境/System Environment：  Intel CPU Ubuntu 20.04
- 版本号/Version：Paddle：  PaddleOCR： 问题相关组件/Related components： paddleOCR 2.5.0.3

First of all, thank you very much for developing such good software.

I am running PaddleOCR in CPU mode, and whenever I perform inference, there is a problem that the memory usage keeps increasing. Is there any way to limit memory usage?

## Paddle OCR Initialization Code
```python
paddle_ch = PaddleOCR(
            show_log=False,
            lang="ch",
            cpu_threads=1,
            det_db_thresh=0.1,
            det_db_box_thresh=0.1,
            use_mp=True,
            enable_mkldnn=True,
            total_process_num=os.cpu_count() * 2 - 1,
            use_angle_cls=True,
            cls_model_dir="whl/cls/ch_ppocr_mobile_v2.0_cls_infer",
            det_model_dir="whl/det/ch/ch_PP-OCRv3_det_infer",
            rec_model_dir="whl/rec/ch/ch_PP-OCRv3_rec_infer"
    )
```

## Memory Profiler Result of PaddleOCR
```
ocr_1  | Line #    Mem usage    Increment  Occurrences   Line Contents
ocr_1  | =============================================================
ocr_1  |    108                                                         # 2. OCR
ocr_1  |    109    556.8 MiB      0.0 MiB          1                   try:
ocr_1  |    110    556.8 MiB      0.0 MiB           1                       if model.language in self.paddle_ocr:
ocr_1  |    111    938.6 MiB    381.8 MiB          1                           paddle_ocr_result = self.paddle_ocr[model.language].ocr(crop_img, cls=False)
```

andyjpaddle:
Maybe you use the multi process mode, the memory usage will increase, but it should reach a peak when  all process is used.

gavinfaimdata:
I'm looking for a way to limit memory usage as well, on GPU.

Running "python3 /tools/infer/predict_system.py" takes about 4.4GB of RAM (on Jetson). I'm using SVTR tiny for rec and mobilenetv3 DB for det.

Issue #4452, opened 2021-10-26
Parallel calls to paddleOCR result in OOM


aruroyc:
Parallel calls to '.ocr' function in paddle_ocr result in OOM even when the ocr object is initialized once.
Code to replicate:
paddle_ocr = PaddleOCR(lang='en',use_gpu=False)
def do_ocr(image):
    paddle_lines = paddle_ocr.ocr(image)
    paddle_lines = [line[1][0] for line in paddle_lines]
    print(paddle_lines)

//Called via endpoint
image= <some image>
do_ocr(image)

Here the memory size increases linearly with the no of parallel http calls even though paddle_ocr was initialized at the module level.

version: tested on both 2.1.0 and 2.2.0

PR #10688, opened 2023-08-21
fix: release memory after predict

DINHDUCCHUNG:
# What?
I have the problem running OCR with CPU/GPU that I run through both AWS EC2 instance and my local machine (CPU) and see memory increase infinitely.
There are some issues like me such as:
- https://github.com/PaddlePaddle/PaddleOCR/issues/7823
- https://github.com/PaddlePaddle/PaddleOCR/issues/4452
- https://github.com/PaddlePaddle/PaddleOCR/issues/7155

# How? 
After tracing, I see `predict_rec.py` not release memory when predicted. I add `self.predictor.try_shrink_memory()` like in `predict_cls.py` and it works.

### DIFF
--- a/tools/infer/predict_rec.py
+++ b/tools/infer/predict_rec.py
@@ -525,7 +525,6 @@ class TextRecognizer(object):
                     if self.benchmark:
                         self.autolog.times.stamp()
                     preds = {"predict": outputs[2]}
-                    self.predictor.try_shrink_memory()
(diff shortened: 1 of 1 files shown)


tink2123:
Thanks for your contribution， please cherry-pick to dygraph branch.

tink2123 (approved):
LGTM

PR #10181, opened 2023-06-15
Fix fitz camelCase deprecation and .PDF not being recognized as pdf file

itasli:

PR #10708, opened 2023-08-23
Cherrypicking GH-10181 to PaddleOCR:dygraph

itasli:
…ile (#10181)

* Fix fitz camelCase deprecation and .PDF not being recognized as pdf file

* refactor get_image_file_list function


shiyutang (approved):
lgtm

PR #10843, opened 2023-09-07
Add tipc for "ParseQ" method

ToddBear:


leduy-it:
Hello, when i try to finetune parseq model, i run inference and get a lot of nan values, i think decode part in this algorithm need to check, thank you for update new method in rec module.

tink2123 (approved):
LGTM

shiyutang:
@leduy-it can you show us the inference script and error image?

PR #10892, opened 2023-09-13
rec_r45_abinet for export model

xlg-go:
When exporting the inference model for ABINet, adapt to the 'image_shape' of ABINetRecResizeImg.


andyjpaddle (approved):
LGTM

PR #11269, opened 2023-11-17
Bug fix for illegal syntax in text detection

travisddavies:
### 描述 Description
In paddleocr.py, there is a bug where it uses pythonic syntax to check if a list is empty, however the dt_boxes for just text detection section uses a numpy array while the text recognition and text detection section above uses a python list. As a result this error is made:
```
----> 2 result = ocr.ocr(num_plate, rec=False, det=True)
      3 for idx in range(len(result)):
      4     res = result[idx]

    672 img = preprocess_image(img)
    673 dt_boxes, elapse = self.text_detector(img)
--> 674 if not dt_boxes:
    675     ocr_res.append(None)
    676     continue

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
To make the format for both the same and to fix the above bug, I changed the pythonic syntax to accommodate for the np array in the ocr function in paddleocr.py

after this fix has been made, calling the following function works: 
```
result = ocr.ocr(num_plate, rec=False, det=True)
```

Edit:
I originally tried changing the dt_box for the detect section to python lists but that creates more bugs, so I simply changed the if statement to accommodate for the numpy array

### 提PR之前的检查 Check-list

- [x] 这个 PR 是提交到dygraph分支或者是一个cherry-pick，否则请先提交到dygarph分支。
- [x] 这个PR已经经过本地测试。This PR can be convered by current tests or already test locally by you.


shiyutang:
Thank you for your contribution, but your PR should only contain your changes and have no conflict with the current branch. please put forward another PR that satisfies the above requirements.

PR #11646, opened 2024-02-29
fix dataset problems release 2.7

yiakwy-xpu-ml-framework-team:
### PR 类型 PR types

Bug fixes

### PR 变化内容类型 PR changes

Dataset, preprocessing :

- remove duplicated data loads of simple dataset
- handling detection task for CTW dataset where annotation polyline data contains more than 4 points (otherwise, resulting forever loop in dataset)
- handling inChannels scalars 
- fasten data loading
- improve model loading

### 描述 Description

improve codes stability and fix bugs when training with cn_PP_OCR_v4 detection model.

I identified failures of dataset infinite loading problem when reproducing a OCR dec in local env. My team is using this model to generate high quality data for Multi-Modal LLM. After this fix , dec can be trained very quickly for a few epochs :

截屏2024-02-29 15 09 45

**SimpleDataset forever loop** : 

> for dataset such as **ctw1500**, each annotation of points of which contains more than 4 points to detect location of arbitrary text in the image. The points may be of the shape **[2\/*the number of annotation in this image\*/, 14\/*number of points\*/, 2\/*2d coordinates\*/]**.

The utility function should also be adapted to process the annotation data instead of throwing them away or raising an exception.

Note by default, "CopyPaste" for **imgaug** method is chosen. That means arbitrary of two images are selected and paired. However get_ext_data always read a new image from the dataset, which is very inefficient.

### 提PR之前的检查 Check-list

- [ ] 这个 PR 是提交到dygraph分支或者是一个cherry-pick，否则请先提交到dygarph分支。
- [x] 这个PR已经经过本地测试。This PR can be covered by existing tests or locally verified.

### DIFF
--- a/ppocr/data/simple_dataset.py
+++ b/ppocr/data/simple_dataset.py
@@ -20,7 +20,7 @@ import random
 import traceback
 from paddle.io import Dataset
 from .imaug import transform, create_operators
-import threading
+
 
 class SimpleDataSet(Dataset):
     def __init__(self, config, mode, logger, seed=None):
@@ -109,7 +109,6 @@ class SimpleDataSet(Dataset):
 
     def get_ext_data(self):
         ext_data_num = 0
-        read_count = 0
         for op in self.ops:
             if hasattr(op, 'ext_data_num'):
                 ext_data_num = getattr(op, 'ext_data_num')
@@ -131,29 +130,20 @@ class SimpleDataSet(Dataset):
             
             if not os.path.exists(img_path):
                 continue
-            # TODO (yiakwy) : cache img_path
             if data['img_path'] in self.img_cache:
                 img = self.img_cache[data['img_path']]
                 # used by imgaug transform
                 data['image'] = img
-                if (read_count + len(self.data_lines)) % 100 == 0:
-                    # pp data loader init a threads pool iterate over dataset
-                    print(f"[SimpleDataset] [{threading.current_thread().name}] fast loaded {read_count}th img {data['img_path']}. ext_data/ext_data_num={len(ext_data)}/{ext_data_num}")
             else:
                 with open(data['img_path'], 'rb') as f:
                     img = f.read()
                     data['image'] = img
                 self.img_cache[data['img_path']] = img
-                if (read_count + len(self.data_lines)) % 100 == 0:
-                    print(f"[SimpleDataset] [{threading.current_thread().name}] {read_count}th img {data['img_path']} readed.")
             if data['img_path'] in self.data_cache:
                 data = self.data_cache[data['img_path']]
             else:
                 data = transform(data, load_data_ops)
                 self.data_cache[data['img_path']] = data
-
-            # read ith image
-            read_count += 1
             
             if data is None:
                 continue
--- a/ppocr/data/simple_dataset.py
+++ b/ppocr/data/simple_dataset.py
@@ -20,7 +20,7 @@ import random
 import traceback
 from paddle.io import Dataset
 from .imaug import transform, create_operators
-
+import threading
 
 class SimpleDataSet(Dataset):
     def __init__(self, config, mode, logger, seed=None):
@@ -57,6 +57,9 @@ class SimpleDataSet(Dataset):
         self.ext_op_transform_idx = dataset_config.get("ext_op_transform_idx",
                                                        2)
         self.need_reset = True in [x < 1 for x in ratio_list]
+        
+        self.img_cache = {}
+        self.data_cache = {}
 
     def set_epoch_as_seed(self, seed, dataset_config):
         if self.mode == 'train':
@@ -106,6 +109,7 @@ class SimpleDataSet(Dataset):
 
     def get_ext_data(self):
         ext_data_num = 0
+        read_count = 0
         for op in self.ops:
             if hasattr(op, 'ext_data_num'):
                 ext_data_num = getattr(op, 'ext_data_num')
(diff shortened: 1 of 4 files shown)


ppocr/data/simple_dataset.py
```diff
@@ -124,18 +128,35 @@ def get_ext_data(self):
...
+                self.data_cache[data['img_path']] = data
 
+            # read ith image
+            read_count += 1
+            
             if data is None:
                 continue
-            if 'polys' in data.keys():
-                if data['polys'].shape[1] != 4:
-                    continue
```
tink2123, line 138: need to handle this situation

ppocr/data/simple_dataset.py
```diff
@@ -124,18 +128,35 @@ def get_ext_data(self):
...
-                img = f.read()
+            # TODO (yiakwy) : cache img_path
+            if data['img_path'] in self.img_cache:
+                img = self.img_cache[data['img_path']]
+                # used by imgaug transform
                 data['image'] = img
-            data = transform(data, load_data_ops)
+                if (read_count + len(self.data_lines)) % 100 == 0:
+                    # pp data loader init a threads pool iterate over dataset
+                    print(f"[SimpleDataset] [{threading.current_thread().name}] fast loaded {read_count}th img {data['img_path']}. ext_data/ext_data_num={len(ext_data)}/{ext_data_num}")
```
tink2123, line 141: The log format please follow the Paddle format.

 self.logger.info(xxx)

tink2123: Can you add a new params to control it without modifying the original code?

ppocr/data/simple_dataset.py
```diff
@@ -124,18 +128,35 @@ def get_ext_data(self):
...
                 continue
-            with open(data['img_path'], 'rb') as f:
-                img = f.read()
+            # TODO (yiakwy) : cache img_path
+            if data['img_path'] in self.img_cache:
+                img = self.img_cache[data['img_path']]
+                # used by imgaug transform
                 data['image'] = img
-            data = transform(data, load_data_ops)
+                if (read_count + len(self.data_lines)) % 100 == 0:
```
tink2123, line 139: Can the parameter "100" be controlled using a configuration file?

ppocr/utils/network.py
```diff
@@ -75,6 +75,9 @@ def maybe_download_params(model_path):
     else:
         url = model_path
     tmp_path = os.path.join(MODELS_DIR, url.split('/')[-1])
+    if os.path.exists(tmp_path) and os.path.isfile(tmp_path):
```
tink2123, line 78: can be deleted

tools/infer/utility.py
```diff
@@ -609,8 +610,31 @@ def get_rotate_crop_image(img, points):
     img_crop = img[top:bottom, left:right, :].copy()
     points[:, 0] = points[:, 0] - left
     points[:, 1] = points[:, 1] - top
-    '''
-    assert len(points) == 4, "shape of points must be 4*2"
```
tink2123, line 613: Situations where more data formats need to be processed in addition to CTW

tink2123 (commented):
Thank you for your contribution！

Your acceleration solution is effective in some scenes. Have you considered adding a parameter to the configuration file to select this way ?

Also, can you please submit a PR to the **dygraph** branch. This modification will be merged into the next stable branch.

ppocr/utils/network.py
yiakwy-xpu-ml-framework-team: Hi models from **Model_Path** was downloaded to tmp_path but it is repeated to do so each time when we launch a task in the same machine, which takes about 3-5 minutes which is unnecessary.

A better way I think is, by providing MD5 sum in model configuration file, we can check if the files in tmp_path is correct.

ppocr/data/simple_dataset.py
yiakwy-xpu-ml-framework-team: Hi, this is where we need discuss. 

This condition leaded to **endless loop** and discarded every annotation line in detection task with points more than 4.

I have checked annotation files we generated and official dataset from ctw1500 (800M) and ICDAR2015 (80M) datasets which are used for finetuning and evaluation: they both have annotation lines with more than 4 points.

The root cause is utility file can only process polygon with fixed points:

> assert len(points) == 4, "shape of points must be 4*2"

Hence we deleted and training goes smoothly again.

tools/infer/utility.py
yiakwy-xpu-ml-framework-team: Yes. But both in CTW, ICDAR2015 and our own annotation files we don't assume bbox consists of fixed number of points to form a polyline.

So we just treat it as general polyline: Not 2 points (upper left, bottom right), 3 points (trianguulars), polygons

yiakwy-xpu-ml-framework-team:
B.t.w I just found this line when I scan the dataset file:

> resized_image = resized_image.astype('float32')

The image can be effectively worked with uint8, this allows 8-bit IO workload and better compute-IO overlap. 

It seems that we havn't use DALI to accelerate IO by allowing device prefetching and on-chip preprocessing.

jzhang533:
Can we land this one ? @tink2123

jzhang533:
@yiakwy-xpu-ml-framework-team you'll need to create a PR, target for dygraph branch. we decided not merging this into release/2.7.

yiakwy-xpu-ml-framework-team:
> @yiakwy-xpu-ml-framework-team you'll need to create a PR, target for dygraph branch. we decided not merging this into release/2.7.

I guess I can cherry-pick this commit onto the dygraph branch. I will do it soon.

PR #11871, opened 2024-04-03
Fix the freezing issue when drawing rectangular annotation boxes on Mac.

thomasbergersen:
Fix the freezing issue when drawing rectangular annotation boxes on Mac.

### DIFF
--- a/PPOCRLabel/libs/canvas.py
+++ b/PPOCRLabel/libs/canvas.py
@@ -13,7 +13,7 @@
 
 import copy
 
-from PyQt5.QtCore import Qt, pyqtSignal, QPointF, QPoint
+from PyQt5.QtCore import Qt, pyqtSignal, QPointF, QPoint, QRectF
 from PyQt5.QtGui import QPainter, QBrush, QColor, QPixmap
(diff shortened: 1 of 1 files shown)


PPOCRLabel/libs/canvas.py
```diff
@@ -593,7 +593,8 @@ def paintEvent(self, event):
             p.setPen(self.drawingRectColor)
             brush = QBrush(Qt.BDiagPattern)
             p.setBrush(brush)
-            p.drawRect(leftTop.x(), leftTop.y(), rectWidth, rectHeight)
+            rect = QRectF(leftTop.x(), leftTop.y(), rectWidth, rectHeight)
```
GreatV, line 596: Why we need `QRectF` Instead of `(x1, y1, w, h)`,  is it a problem with the `int` type?

thomasbergersen: Yes ,Type conversion problem. 
Mac OS v13.6.6  ,Python 3.10.

thomasbergersen:
> Would you mind cherry-pick the changes to the main branch?

Okay, update later.

PR #11887, opened 2024-04-08
handle triangular contour exceptions

zovelsanj:
In case any two coordinates of a contour are same, it will make a triangular region instead of rectangular one. Thus, when creating polygon from that contour, the area of the box (and thus distance) will be zero, resulting in an empty box. This results the following error during minimum area calculation:
```
bounding_box = cv2.minAreaRect(contour)
cv2.error: OpenCV(4.6.0) D:\a\opencv-python\opencv-python\opencv\modules\imgproc\src\convhull.cpp:143: error: (-215:Assertion failed) total >= 0 && (depth == CV_32F || depth == CV_32S) in function 'cv::convexHull'
```

### 提PR之前的检查 Check-list

- [ ] 这个 PR 是提交到dygraph分支或者是一个cherry-pick，否则请先提交到dygarph分支。
- [ ] 这个PR清楚描述了功能，帮助评审能提升效率。This PR have fully described what it does such that reviewers can speedup.
- [ ] 这个PR已经经过本地测试。This PR can be covered by existing tests or locally verified.


jzhang533:
I think the fix is correct, but we changed the dev workflow, this patch should be applied to main branch.

zovelsanj:
> I think the fix is correct, but we changed the dev workflow, this patch should be applied to main branch.

Thanks @jzhang533, I have now changed the base to main.

PR #11916, opened 2024-04-11
Enhance the OCR recognition accuracy of PPStructure.

RussellLuo:
This PR is likely to close the following issues:
- #10270
- #10322
- #10367
- #11278
- #11665
- #11302
- #11328 
- #11869

### DIFF
--- a/ppstructure/predict_system.py
+++ b/ppstructure/predict_system.py
@@ -58,7 +58,6 @@ class StructureSystem(object):
                 logger.warning(
                     "When args.layout is false, args.ocr is automatically set to false"
                 )
-            args.drop_score = 0
             # init model
             self.layout_predictor = None
             self.text_system = None
@@ -93,6 +92,7 @@ class StructureSystem(object):
             'all': 0
         }
         start = time.time()
+
         if self.image_orientation_predictor is not None:
             tic = time.time()
             cls_result = self.image_orientation_predictor.predict(
@@ -108,6 +108,7 @@ class StructureSystem(object):
                 img = cv2.rotate(img, cv_rotate_code[angle])
             toc = time.time()
             time_dict['image_orientation'] = toc - tic
+
         if self.mode == 'structure':
             ori_im = img.copy()
             if self.layout_predictor is not None:
@@ -116,6 +117,20 @@ class StructureSystem(object):
             else:
                 h, w = ori_im.shape[:2]
                 layout_res = [dict(bbox=None, label='table')]
+
+            # As reported in issues such as #10270 and #11665, the old
+            # implementation, which recognizes texts from the layout regions,
+            # has problems with OCR recognition accuracy.
+            #
+            # To enhance the OCR recognition accuracy, we implement a patch fix
+            # that first use text_system to detect and recognize all text information
+            # and then filter out relevant texts according to the layout regions.
+            text_res = None
+            if self.text_system is not None:
+                text_res, ocr_time_dict = self._predict_text(img)
+                time_dict['det'] += ocr_time_dict['det']
+                time_dict['rec'] += ocr_time_dict['rec']
+
             res_list = []
             for region in layout_res:
                 res = ''
@@ -126,6 +141,8 @@ class StructureSystem(object):
                 else:
                     x1, y1, x2, y2 = 0, 0, w, h
                     roi_img = ori_im
+                bbox = [x1, y1, x2, y2]
+
                 if region['label'] == 'table':
                     if self.table_system is not None:
                         res, table_time_dict = self.table_system(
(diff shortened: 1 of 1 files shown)


__init__.py
```diff
@@ -15,7 +15,7 @@
 
 __version__ = paddleocr.VERSION
 __all__ = [
-    'PaddleOCR', 'PPStructure', 'draw_ocr', 'draw_structure_result',
+    'PaddleOCR', 'PPStructure', 'PPStructureX', 'draw_ocr', 'draw_structure_result',
```
GreatV, line 18: Can `PPStructure` be modified directly instead of creating a new class? I prefer to have only one `PPStructure`.

RussellLuo: Agreed, I'll got it done soon!

GreatV:
For the picture below, this hot fix didn't work.

INPUT：

test_1

OUTPUT ：

result

RussellLuo:
In my own opinion, this fix is essentially just a patch (rather than a complete solution), so there may be edge cases where this fix does not address.

__init__.py
RussellLuo: Done.

RussellLuo:
I took a screenshot of your INPUT image (intentionally leaving extra white space around the last line), and got the OUTPUT as below:

image

(I guess the omission of the last line of text is due to the original image being cropped.)

GreatV:
https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L120-L149

Could you modify the code here to make it detect and recognize the original image directly, so that the other interfaces don't need to be changed?

RussellLuo:
I'm not familiar with this piece of code. To my understanding, the core logic is to first detect the layout regions (by using the LayoutPredictor):

https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L114

and then recognize the corresponding texts from each layout region (by using the TextSystem):

https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L120-L125

As shown above, to get the image pixels `roi_img` within the layout region, the original image `ori_im` is cropped according to the coordinates of the layout region. **As discussed in #10270, this exact operation is suspected to affect the precision of OCR recognition.**

Based on the analysis, the core idea behind this fix is to:

1. First detect all possible text regions and recognize the texts within them (by only using PaddleOCR, which is essentially a TextSystem).
2. Then collect texts from the text regions that intersect with each layout region (detected by using a StructureSystem).

Therefore,  this fix is a hybrid solution that leverages both PaddleOCR and StructureSystem, but I'm not sure whether  it is appropriate to place the changes of this hybrid solution into [PaddleOCR/ppstructure/predict_system.py](https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L120-L149).

Hope to hear your suggestions.

GreatV:
> Based on the analysis, the core idea behind this fix is to:

> 1. First detect all possible text regions and recognize the texts within them (by only using PaddleOCR, which is essentially a  TextSystem).
>2. Then collect texts from the text regions that intersect with each layout region (detected by using a StructureSystem).

Yes, I think that would be a more appropriate modification, with minimal impact on the overall code structure. I hope you'll give it a try.

ppstructure/predict_system.py
```diff
@@ -116,6 +117,20 @@ def __call__(self, img, return_ocr_result_in_table=False, img_idx=0):
...
                 layout_res = [dict(bbox=None, label='table')]
+
+            # As reported in issues such as #10270 and #11665, the old
+            # implementation, which recognizes texts from the layout regions,
+            # has problems with OCR recognition accuracy.
+            #
+            # To enhance the OCR recognition accuracy, we implement a patch fix
+            # that first detect all text regions by using the text_detector
+            # and then recognize the texts from the text regions (intersecting
+            # with the layout regions) by using the text_recognizer.
```
GreatV, line 128: I prefer to use `text_system` to first detect and recognize all text information, and then filter out relevant text based on the layout information.

paddleocr.py
```diff
@@ -637,7 +637,7 @@ def __init__(self, **kwargs):
     def ocr(self, img, det=True, rec=True, cls=True, bin=False, inv=False, alpha_color=(255, 255, 255)):
         """
         OCR with PaddleOCR
-        
+
```
GreatV, line 640: These may not need to be changed.

tools/infer/predict_system.py
```diff
@@ -64,7 +64,7 @@ def draw_crop_rec_res(self, output_dir, img_crop_list, rec_res):
             logger.debug(f"{bno}, {rec_res[bno]}")
         self.crop_image_res_index += bbox_num
 
-    def __call__(self, img, cls=True):
+    def __call__(self, img, cls=True, dt_boxes=None):
```
GreatV, line 67: `dt_boxes`  was not used.

ppstructure/predict_system.py
RussellLuo: I'll give this a try.

RussellLuo: One more question here, the following recover part has not been changed:

https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L139-L143

Should we use the same mechanism for this piece of logic?

GreatV: It would be advisable to update all the code in this section to match the current logic as well.

RussellLuo:
All changes have been centralized in `StructureSystem`, and here are some additional notes.

The following line would keep results with low confidence and has been deleted:
https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L61

These two lines would cause coordinate error and have been commented out. It took me a significant amount of time to identify this issue. I don't understand the logic behind this piece of code, perhaps these lines should be deleted?
https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L165-L166

GreatV:
Why do we need to keep low confidence objectives?

https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L61

The code here is used to calculate the offset from the results of the layout analysis and can be removed.

https://github.com/PaddlePaddle/PaddleOCR/blob/c82dd6406eecb3c0840b26b055b436a883128307/ppstructure/predict_system.py#L165-L166

RussellLuo:
> Why do we need to keep low confidence objectives?

This line is from the previous code and will lead to the return of results with low confidence (as the default value `0.5` was reset). The The latest change has been made to remove this line.

> The code here is used to calculate the offset from the results of the layout analysis and can be removed.

Got it!

RussellLuo:
All suggested changes have been made and the all commits have been squashed into one.

GreatV:
@RussellLuo  Thanks for your contribution, I will take some time to check it again.

GreatV:
old:
result_old

new:
result_new

RussellLuo:
@GreatV Thanks for your patient review and excellent suggestions!

jzhang533:
thanks, lets merge it.

luotao1:
@RussellLuo Thanks for your contribution! You will receive a beautiful PaddlePaddle gift. Please provide your mailing address by filling out the following questionnaire before October 18th.

luotao1:
hi, @RussellLuo 
- 非常感谢你对飞桨的贡献，我们正在运营一个PFCC组织，会通过定期分享技术知识与发布开发者主导任务的形式持续为飞桨做贡献，详情可见 https://github.com/luotao1 主页说明。
- 如果你对PFCC有兴趣，请发送邮件至 [ext_paddle_oss@baidu.com](mailto:ext_paddle_oss@baidu.com)，我们会邀请你加入~

PR #12042, opened 2024-05-02
Save detection inference logs to file

AlexPasqua:
I added the possibility to save the detection logs to file.

The code was almost all there already, just some arguments weren't passed the right way in order to save the detection logs to file.

In case, we could also implement this feature for the other scripts in `tools/infer/`.

### DIFF
--- a/tools/infer/predict_det.py
+++ b/tools/infer/predict_det.py
@@ -32,11 +32,9 @@ from ppocr.data import create_operators, transform
 from ppocr.postprocess import build_post_process
 import json
 
-logger = get_logger()
-
 
 class TextDetector(object):
-    def __init__(self, args):
+    def __init__(self, args, logger):
         self.args = args
         self.det_algorithm = args.det_algorithm
         self.use_onnx = args.use_onnx
@@ -158,7 +156,7 @@ class TextDetector(object):
                 model_precision=args.precision,
                 batch_size=1,
                 data_shape="dynamic",
-                save_path=None,
+                save_path=None,     # not used if logger is not None
                 inference_config=self.config,
                 pids=pid,
                 process_name=None,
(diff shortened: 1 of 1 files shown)


tools/infer/predict_det.py
```diff
@@ -383,10 +381,18 @@ def __call__(self, img):
...
-    text_detector = TextDetector(args)
     total_time = 0
     draw_img_save_dir = args.draw_img_save_dir
     os.makedirs(draw_img_save_dir, exist_ok=True)
+    
+    # logger
+    log_file = args.save_log_path
+    if os.path.isdir(args.save_log_path) or (not os.path.exists(args.save_log_path) and args.save_log_path.endswith("/")):
+        log_file = os.path.join(log_file, "benchmark_detection.log")
+    logger = get_logger(log_file=log_file)
```
jzhang533, line 392: seems this logger should be prepared and passed into AutoLog as a arg in https://github.com/PaddlePaddle/PaddleOCR/blob/main/tools/infer/predict_det.py#L168

I am worrying about maintenance of AutoLog module, it's been implemented in https://github.com/LDOUBLEV/AutoLog, and published to a server as documented here: https://github.com/PaddlePaddle/PaddleOCR/blob/38c0c9ee7715d0600c0c7ef2784b3ce64edd0710/test_tipc/prepare.sh#L204.

I think this is only been used by a terminated [tipc plan](https://github.com/PaddlePaddle/PaddleOCR/tree/main/test_tipc), we should redesign this feature, or deprecate this feature.

AlexPasqua: > I am worrying about maintenance of AutoLog module, it's been implemented in https://github.com/LDOUBLEV/AutoLog, and published to a server as documented here:

In this case PaddleOCR could implement its own logger, instead of using AutoLog.

I think that it's a little bit of a pity to have a feature to run benchmarks without the option to save the results to a file (and only seeing them on screen instead), therefore I think this feature would be very useful.

tools/infer/predict_det.py
```diff
@@ -32,11 +32,9 @@
 from ppocr.postprocess import build_post_process
 import json
 
-logger = get_logger()
-
 
 class TextDetector(object):
-    def __init__(self, args):
+    def __init__(self, args, logger):
```
jzhang533, line 37: logger arg needs a default value None, the default behavior should be using logger from `logger = get_logger()`.

otherwise, it will break this: https://github.com/PaddlePaddle/PaddleOCR/blob/1569cfb430dee06b23413b7b0a6d99dc8c84ab70/tools/infer/predict_system.py#L51

tools/infer/predict_det.py
jzhang533: agree

jzhang533 (approved):
LGTM

luotao1:
@AlexPasqua Thanks for your contribution! You will receive a beautiful PaddlePaddle gift. Please provide your mailing address by filling out the following questionnaire before October 18th.

luotao1:
@AlexPasqua Could you provide your phone number? We need this information when mailing gifts. Please send your phone number to ext_paddle_oss@baidu.com. Thanks very much!

PR #12044, opened 2024-05-03
Corrected model path for ONNX inference

lsch0lz:
Hi!
Both paths don't need the directory before the filename, also one backslash was missing


GreatV:
hi @lsch0lz , CLA is not signed yet.

PR #12142, opened 2024-05-20
fix the issue of repeatedly downloading pretrained model

GreatV:
当配置文件中指定了 `pretrained`，每次训练时都会重复下载模型，而不是先检查是否存在此模型。

https://github.com/PaddlePaddle/PaddleOCR/blob/960243862f6c6833f58e39f3b74864cba15edfeb/configs/det/ch_PP-OCRv4/ch_PP-OCRv4_det_cml.yml#L58-L60

### DIFF
--- a/ppocr/utils/network.py
+++ b/ppocr/utils/network.py
@@ -24,6 +24,8 @@ MODELS_DIR = os.path.expanduser("~/.paddleocr/models/")
 
 
 def download_with_progressbar(url, save_path):
+    if save_path and os.path.exists(save_path):
+        return
     logger = get_logger()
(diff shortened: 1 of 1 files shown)


ppocr/utils/network.py
```diff
@@ -24,6 +24,8 @@
 
 
 def download_with_progressbar(url, save_path):
+    if save_path and os.path.exists(save_path):
+        return
```
jzhang533, line 28: Having an information log before returning might help users understand what's going on.

jzhang533 (approved):
LGTM

PR #12516, opened 2024-05-27
update common pre-commit configs

jzhang533:
update common pre-commit configs and commit the results of running pre-commit run -a

will mark this PR as WIP, since we may need to discuss are there any other checks needed for the repository.

please review .pre-commit-config.yaml file, all others are the results of running `pre-commit run -a`.


.pre-commit-config.yaml
```diff
@@ -31,7 +27,7 @@ repos:
         files: \.(c|cc|cxx|cpp|cu|h|hpp|hxx|cuh|proto)$
 # For Python files
 -   repo: https://github.com/psf/black.git
```
SWHL, line 29: I'm just curious why the Flake8 tool is used for the benchmark and test_tipc directory? 
Why not just use Black tools?

.pre-commit-config.yaml
```diff
@@ -1,26 +1,22 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: a11d9314b22d8f8c7556443875b731ef05965464
+    rev: v4.6.0
     hooks:
+    -   id: check-added-large-files
+        args: ['--maxkb=512']
```
GreatV, line 6: I'm more concerned that 512kb won't be enough for our use, as some of the previous files had more than 512kb in them.

.pre-commit-config.yaml
GreatV: @SWHL  These folders are excluded because they contain numerous bugs that are too time-consuming to fix and will be addressed at a later date.

.pre-commit-config.yaml
jzhang533: This will only be applied to local staged files as documented here: https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#check-added-large-files. which means, current .github/workflows/codestyle.yml will not capture large files being added.

Another option : increase the limit to 1024kb, and check all the files on each pull_request by using `--enforce-all`, at the same time, exclude existing large files. there are 28 files larger than 1M in the repo.

GreatV (approved):
LGTM

PR #12520, opened 2024-05-28
Update pyproject.toml

GreatV:
`setuptools_scm` may conflict with `manifest.in`, disable it for temporary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm", "wheel"]
+requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]


jzhang533 (approved):
LGTM

validated locally by running:
`git clean -fxd && python -m build`

there is a notice in [setuptools_scm](https://github.com/pypa/setuptools_scm?tab=readme-ov-file#about)
> [setuptools-scm](https://github.com/pypa/setuptools_scm) extracts Python package versions from git or hg metadata instead of declaring them as the version argument or in an SCM managed file.

> Additionally, [setuptools-scm](https://github.com/pypa/setuptools_scm) provides setuptools with a list of files that are managed by the SCM
(i.e. it automatically adds all of the SCM-managed files to the sdist).
Unwanted files must be excluded via MANIFEST.in.

PR #12542, opened 2024-05-29
Save recognition inference logs to file

AlexPasqua:
I added the possibility to save the recognition logs to file.

Same as #12042 but for recognition instead of detection.

If later you want to implement your own logger and move away from AutoLog (as briefly mentioned in [this comment](https://github.com/PaddlePaddle/PaddleOCR/pull/12042#discussion_r1601502917)), that's also good, but it will require more changes to the code. So, in the meanwhile, this PR quickly implements logging to file for recognition inference 😄


jzhang533 (approved):
LGTM
thanks for contributing.

PR #13136, opened 2024-06-19
temporarily set numpy < 2.0

GreatV:
We need to remove imgaug, but it will take some time, so we will temporarily set numpy to < 2.0.


jzhang533 (approved):
LGTM

PR #13142, opened 2024-06-20
Update japan_dict.txt

madmalkav:
Update japan_dict.txt to include missing jouyou kanji ( https://github.com/PaddlePaddle/PaddleOCR/discussions/12940 )


GreatV:
Will this affect the previous jp model?

madmalkav:
It is my understanding this file is only used when training a new model, so I don't think it would have any effect until a new model is trained, but there is a chance I misunderstood everything wrong and it works in a different way 😅

GreatV (commented):
Since they use the same character dictionary, using the previous model may result in inconsistent output, and I think it would be better to use a different filename (e.g., ja_ext_dict.txt).

madmalkav:
If this file can indeed affect the current model, I would propose to hold this PR until a new version of the japanese model is going to be trained. I don't like too much the idea of creating a new file and called it "extended" because this is not really extending the dictionary, is fixing a fault in it.

madmalkav:
Can I ask why t[his similar PR](https://github.com/PaddlePaddle/PaddleOCR/pull/11125) for another language was already merged? What is the difference with my PR? I want to understand to see if I can do something else to move this forward.

GreatV:
I don't think that PR being merged will work properly with the previous model.

GreatV:
Maybe we could try to keep the version information of the character dictionary (a new column in the model list is used to show the dictionary used by the model), so that new character dictionaries can be merged in, and old models use old character dictionaries.


| model code|description|character dictionary|model size|download|Update Date|
|:------:|:------|:------|:------:|:------|:------:|
|ch|Chinese and English|https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.7/ppocr/utils/ppocr_keys_v1.txt|3.71M|[inference model](https://paddleocr.bj.bcebos.com/dygraph_v2.0/ch/ch_ppocr_mobile_v2.0_rec_infer.tar)/</br>[trained model](https://paddleocr.bj.bcebos.com/dygraph_v2.0/ch/ch_ppocr_mobile_v2.0_rec_train.tar)|2020.9.22|
|ch|Chinese and English|https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.8/ppocr/utils/ppocr_keys_v1.txt|3M|[inference model](https://paddleocr.bj.bcebos.com/dygraph_v2.0/ch/ch_ppocr_mobile_v2.0_rec_infer.tar)/</br>[trained model](https://paddleocr.bj.bcebos.com/dygraph_v2.0/ch/ch_ppocr_mobile_v2.0_rec_train.tar)|2024.9.22|

thebest132:
@GreatV @madmalkav Is this fixed? PaddleOCR does still seem to mess up some Japanese characters.

madmalkav:
No, it isn’t 🥲

thebest132:
Pinging @jzhang533 too. This is 100% necessary to fix for any sort of production-ready Japanese recognition.

rycont:
Why not yet merged? Is there any problem with this PR? I'm really waiting to use PaddleOCR for a production usage.

test3211234:
@rycont They may be waiting until they train a new model which probably isn't too often. Hopefully they don't forget about this.

GreatV:
@rycont PP-OCRV5 already supports Japanese recognition. You can refer to [text_recognition](https://www.paddleocr.ai/latest/en/version3.x/module_usage/text_recognition.html), and the dictionary file can be found at [ppocr/utils/dict/ppocrv5_dict.txt](https://github.com/PaddlePaddle/PaddleOCR/blob/release/3.2/ppocr/utils/dict/ppocrv5_dict.txt.)

madmalkav:
@GreatV if it is fixed in the new version, why this PR and  issue #12940 are still open and have no comment

GreatV:
This issue has been addressed in PP-OCRv5, which uses a new character dictionary with broader Japanese character coverage. We recommend using the PP-OCRv5 recognition model instead of the older Japanese models.

PR #13188, opened 2024-06-25
Modify the version to 2.6.1.0, because paddleocr cannot be imported from setup.py in some Docker environments.

changdazhou:


cuicheng01 (approved):

UserWangZz (approved):
Fix the issue of importing Paddle version in setup causing unsuccessful installation in certain environments

PR #13389, opened 2024-07-15
line 445 program.py

ManikSinghSarmaal:
refering to-https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/incubate/ModelAverage_en.html

it should be paddle.incubate.ModelAverage and not paddle.incubate.optimizer.ModelAverage

### DIFF
--- a/tools/program.py
+++ b/tools/program.py
@@ -442,7 +442,7 @@ def train(
                 and dist.get_rank() == 0
             ):
                 if model_average:
-                    Model_Average = paddle.incubate.optimizer.ModelAverage(
+                    Model_Average = paddle.incubate.ModelAverage(
                         0.15,
                         parameters=model.parameters(),
                         min_average_window=10000,


GreatV:
Thank you for your contribution, I'm more curious if the previous form of the code leads to training errors?

ManikSinghSarmaal:
> Thank you for your contribution, I'm more curious if the previous form of the code leads to training errors?

I have not used the use_srn flag that is for config["Architecture"]["algorithm"] == "SRN", I was going around the code and checking website at the same time so came across this, I don't think so it will usually lead to any training errors but to be on the safe side always😄

GreatV (approved):
LGTM

luotao1:
@ManikSinghSarmaal Thanks for your contribution! You will receive a beautiful PaddlePaddle gift. Please provide your mailing address by filling out the following questionnaire before October 18th.

Issue #13538, opened 2024-07-29
版面恢复md，有些地方错误


keyfall:
### Search before asking

- [X] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.

- [X] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.

- [X] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

### Bug

https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppstructure/recovery/README_ch.md
上述文档中
```paddleocr --image_dir=ppstructure/recovery/UnrealText.pdf --type=structure --recovery=true --use_pdf2docx_api=true```
总共有两个问题：
1.该命令的--image_dir位置不对，应该是ppstructure/docs/recovery/UnrealText.pdf,
2.使用了这个命令之后，结果显示
![image](https://github.com/user-attachments/assets/d8cf3edb-3a16-40f1-841b-c2af925d9aae)

使用你们提供的命令，与文件，结果出不来，希望修改一下

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

GreatV:
什么环境，什么paddle版本，paddleocr版本

Gmgge:
看了下应该是目录结构调整的时候，忽略了该说明文档中的路径调整，方便提个pr嘛？麻烦在pr中链接下该问题。

PR #13575, opened 2024-08-02
docs: Update the pdf file path in the operation demonstration

Gmgge:
Document path adjustment, update the path in the demonstration，from [issue](https://github.com/PaddlePaddle/PaddleOCR/issues/13538)


GreatV (approved):
LGTM

PR #13698, opened 2024-08-19
Provides Vietnamese dictionary and corpus

lingskr:
I have in the folder "ppocr/utils/dict" I committed the dictionary text to this path and named it "vi_dict.txt" which contains a list of all characters. The Vietnamese dictionary is from Wikipedia.

I did not find the corpus in the folder "ppocr/utils/corpus"; I could only view it in my GitHub. [Link here](https://github.com/lingskr/Vietnamese-Corpus-and-Dictionary).


GreatV (approved):
LGTM

PR #13748, opened 2024-08-26
Add Lao dictionary

lingskr:
I added a lao_dict.txt file to ppocr/utils/dict/


GreatV:
Please do not create the same PR multiple times, unless it is truly necessary.

PR #13918, opened 2024-09-27
fix hubserving run error

GreatV:
fix
- https://github.com/PaddlePaddle/PaddleOCR/pull/13896#issuecomment-2378523045

### DIFF
--- a/paddleocr.py
+++ b/paddleocr.py
@@ -915,8 +915,10 @@ def main():
     """
     # for cmd
     args = parse_args(mMain=True)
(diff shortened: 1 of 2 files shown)


paddleocr.py
```diff
@@ -915,8 +915,10 @@ def main():
     """
     # for cmd
     args = parse_args(mMain=True)
+    logger.info("for uage help, please use `paddleocr --help`")
```
GreatV, line 918:
```suggestion
    logger.info("for usage help, please use `paddleocr --help`")
```

PR #14000, opened 2024-10-15
Update predict_det.py

deisler134:
"order_points_clockwise" could cause some bugs when points have more than one max values


GreatV:
Can you write some unit tests to verify the effectiveness of the changes? Or give some examples of wrong case.

PR #14142, opened 2024-11-01
Support inference for GCU

EnflameGCU:
配置`GCU`设备的推理流程

### DIFF
--- a/tools/infer/utility.py
+++ b/tools/infer/utility.py
@@ -41,6 +41,12 @@ def init_args():
     parser.add_argument("--use_xpu", type=str2bool, default=False)
     parser.add_argument("--use_npu", type=str2bool, default=False)
     parser.add_argument("--use_mlu", type=str2bool, default=False)
+    parser.add_argument(
+        "--use_gcu",
+        type=str2bool,
+        default=False,
+        help="Use Enflame GCU(General Compute Unit)",
+    )
     parser.add_argument("--ir_optim", type=str2bool, default=True)
     parser.add_argument("--use_tensorrt", type=str2bool, default=False)
     parser.add_argument("--min_subgraph_size", type=int, default=15)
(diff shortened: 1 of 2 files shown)


tools/infer/utility.py
```diff
@@ -25,6 +25,8 @@
 import random
 from ppocr.utils.logging import get_logger
 
+import paddle_custom_device.gcu.passes as gcu_passes
```
GreatV, line 28: Regular users don't really need `paddle_custom_device`.

EnflameGCU: You are right, this needs to be corrected. Thank you for the correction.

tools/infer/utility.py
```diff
@@ -41,6 +41,7 @@ def init_args():
     parser.add_argument("--use_xpu", type=str2bool, default=False)
     parser.add_argument("--use_npu", type=str2bool, default=False)
     parser.add_argument("--use_mlu", type=str2bool, default=False)
+    parser.add_argument("--use_gcu", type=str2bool, default=False)
```
jzhang533, line 44: please include a short message to help user understand its purpose.
To me, at the first glance, I am guessing gcu stands for GraphCore Unit, realized it's Enflame's device after searching. 

Other arguments(i.e.: uie_xpu, use_npu, use_mlu) also need help message to clarify the respective devices, but could be addressed in separate PRs.

EnflameGCU: OK, thanks for pointing that out. I'll add comments to GCU first. 

I am afraid that our description of other arguments may not be accurate, and they may need to be described by the developers of the corresponding hardware manufacturers.

tools/infer/utility.py
```diff
@@ -41,6 +41,9 @@ def init_args():
     parser.add_argument("--use_xpu", type=str2bool, default=False)
     parser.add_argument("--use_npu", type=str2bool, default=False)
     parser.add_argument("--use_mlu", type=str2bool, default=False)
+    parser.add_argument(
+        "--use_gcu", type=str2bool, default=False
+    )  # Use Enflame GCU(General Compute Unit)
```
jzhang533, line 46:
```suggestion
    parser.add_argument(
        "--use_gcu", type=str2bool, default=False,
        help="Use Enflame GCU(General Compute Unit)",
    )
```

tools/infer/utility.py
jzhang533: > I am afraid that our description of other arguments may not be accurate, and they may need to be described by the developers of the corresponding hardware manufacturers.

agree

tools/infer/utility.py
EnflameGCU: Thank you for your suggestion.

tools/infer/utility.py
```diff
@@ -293,6 +299,28 @@ def create_predictor(args, mode, logger):
             config.enable_custom_device("mlu")
         elif args.use_xpu:
             config.enable_xpu(10 * 1024 * 1024)
+        elif args.use_gcu:  # for Enflame GCU(General Compute Unit)
+            import paddle_custom_device.gcu.passes as gcu_passes
```
cuicheng01, line 303: Should we first check whether the `paddle_custom_device` package is present? If it’s not, remind the user to install it first?

tools/infer/utility.py
```diff
@@ -293,6 +299,28 @@ def create_predictor(args, mode, logger):
...
+                    "gcu", 0, paddle.inference.PrecisionType.Half
+                )
+                gcu_passes.set_exp_enable_mixed_precision_ops(config)
+            else:
+                config.enable_custom_device("gcu")
+
+            if paddle.framework.use_pir_api():
+                config.enable_new_ir(True)
+                config.enable_new_executor(True)
+                kPirGcuPasses = gcu_passes.inference_passes(
```
nepeplwu, line 317: Why do we need to build passes based on "PaddleOCR" or "PaddleXXX"?

tools/infer/utility.py
EnflameGCU: Thank you for your suggestion.

tools/infer/utility.py
EnflameGCU: We need to optimize performance based on hardware features and even network structure, and the processing here is to increase this flexibility.

cuicheng01 (approved):

PR #14147, opened 2024-11-02
Retrieved and returned detection confidence score

AlexPasqua:
There might still be something to do with `sub_dt_scores`. I cannot throughly test on this current machine, take this PR as an initial draft, even though it might be almost ready.

I look forward to your feedback!


GreatV (commented):
Hi @AlexPasqua, please update unit tests to fix CI.

PR #14202, opened 2024-11-11
Remove Python 2 compatibility dependency six

cclauss:
https://pypi.org/project/six is no longer needed.

https://github.com/PaddlePaddle/PaddleOCR/blob/9308d5e7884e322008b2cd5ee7e23e02a254a1ac/pyproject.toml#L17


jzhang533 (approved):
LGTM
thanks for doing this.

PR #14418, opened 2024-12-18
Return separate Excel sheets for tables while using paddleocr structure 

SHUBH4M-KUMAR:
modified predict.py file to return separate excel sheet according to the number of tables detected.


GreatV:
please resubmit this PR to main branch

PR #14571, opened 2025-01-20
Fix language error and spelling mistakes in the documentation

timminator:
I think I found a mistake in your documentation regarding the supported languages.
You are stating that you support the language "Saudi Arabia" with the language code "sa".
I think you mean "Sanskrit" in this case.
When looking at the PaddleOCR.py file, "sa" is also sorted in the devanagari list, which supports my assumption. And "Saudi Arabia" is also not a language at all.
While I was at it I also corrected a few spelling mistakes I noticed.

### DIFF
--- a/docs/ppocr/blog/multi_languages.en.md
+++ b/docs/ppocr/blog/multi_languages.en.md
@@ -187,17 +187,17 @@ If necessary, you can read related documents:
 |English|en| |Hindi|hi|
 |French|fr| |Uyghur|ug|
 |German|german| |Persian|fa|
-|Japan|japan| |Urdu|ur|
+|Japanese|japan| |Urdu|ur|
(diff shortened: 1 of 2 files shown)


GreatV:
You’re correct. From the code,

https://github.com/PaddlePaddle/PaddleOCR/blob/cf4c0591ba475e3560d6a596fa3b3109facd0ee1/paddleocr.py#L487-L501

It seems "sa" refers to a Devanagari language, most likely Sanskrit.

luotao1:
@timminator Thanks for your contribution! You will receive a beautiful PaddlePaddle gift. Please fill out the following questionnaire before **June 6th** to provide your mailing address.

Looking forward to the future, we will walk further together in the open source world!
Click Here：https://paddle.wjx.cn/vm/ws1JZNP.aspx#

PR #14621, opened 2025-02-05
CPP: tidied file header includes

nonwill:
Fix header including to avoid confusions: headers are only included where they are needed. 

Sources compiled OK with both mingw64 + gcc-14 and msvc-2019, and the msvc-2019 build of ppocr.exe works well.


GreatV (approved):
LGTM

PR #14641, opened 2025-02-08
Add english docu for CMAKE compilation on windows

timminator:
Hi!

Last week I wanted to try out the C++ inference version of PaddleOCR and noticed that the compilation guide for windows only existed in chinese and it was also quite outdated aswell.

So after I got it working and compiled paddleocr successfully on windows I took the time to write an english cmake compilation guide and updated it with my recent findings.

I've also changed the link in the cpp infer readme to link now to the english version that I created because it itself was written in english. On top of the readme of both CMAKE compilation guides I added buttons to switch between the languages.

I hope you like it!

### DIFF
--- a/deploy/cpp_infer/docs/windows_vs2019_build.md
+++ b/deploy/cpp_infer/docs/windows_vs2019_build.md
@@ -1,3 +1,5 @@
+[English](windows_vs2022_build.md) | 简体中文
+
 - [Visual Studio 2019 Community CMake 编译指南](#visual-studio-2019-community-cmake-编译指南)
   - [1. 环境准备](#1-环境准备)
     - [1.1 安装必须环境](#11-安装必须环境)
--- a/deploy/cpp_infer/docs/windows_vs2022_build.md
+++ b/deploy/cpp_infer/docs/windows_vs2022_build.md
@@ -1,6 +1,4 @@
----
-comments: true
----
+English | [简体中文](windows_vs2019_build.md)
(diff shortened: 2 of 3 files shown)


GreatV (commented):
We have moved the documents to a dedicated `docs` directory so that we can view them more conveniently on the documentation site, such as [windows_vs2019_build](https://paddlepaddle.github.io/PaddleOCR/latest/en/ppocr/infer_deploy/windows_vs2019_build.html), with the corresponding source link being [windows_vs2019_build.md](https://github.com/PaddlePaddle/PaddleOCR/edit/main/docs/ppocr/infer_deploy/windows_vs2019_build.en.md). It would be even better if you could update and modify the corresponding documents in the `docs` directory as well.

docs/ppocr/infer_deploy/windows_vs2019_build.en.md
GreatV, line 1: Please rename the file to `windows_vs2022_build.en.md`

timminator: Noticed that aswell. Updated it again.
I kept the old name vs2019_build.en in the docs even though its now for 2022 because if it would have a different name than the chinese version I think switching between languages on the website would break right?
And I have no knowledge of chinese, so I am not able to update the chinese guide aswell.

GreatV (approved):
LGTM

PR #14686, opened 2025-02-14
use the env variable PADDLE_OCR_BASE_DIR if it exists to download models

mauryaland:
Sorry for the second PR on this topic, I noticed afterward that another path (`MODELS_DIR`) was also hardcoded for downloading models. It will now use the environment variable  `PADDLE_OCR_BASE_DIR` if this one exists.


GreatV (approved):
LGTM

PR #14819, opened 2025-03-07
Fix wrong date in newly added capabilities

timminator:
The last commit that described the newly added capabilities listed the previous year. ;-)

I've updated this to 2025.


GreatV (approved):
LGTM

PR #14885, opened 2025-03-17
docs: Fix formatting of a Markdown List

jk4e:
This should fix the display of the list.

Currently, it looks like this:

image

Should look like this:

image

Thanks! 🚀


GreatV (approved):
LGTM

PR #15025, opened 2025-04-16
Fixed deduplication in picodet_postprocess

trueMiskin:
Hi,

When I tested your OCR with layout analysis I ran into a weird bug. I had this image:
Input image
Firstly, I ran the model with a score threshold 0.4 and the output was this image:
result_0 4
I was a little bit upset that the model did not find the title with high confidence so I lowered the threshold to 0.3 and the result is this image:
result_0 3
When I saw it, I knew that something was going wrong. When the threshold is lowered, the same or more bboxes can appear, but they cannot disappear.

I thought there was some bug in the nms function but I found a bug in the deduplication function. The function iterates bboxes and computes iou with all other bboxes. If there are overlapping bboxes one persists and the other is removed (marked as duplicate) but this approach is wrong. For example in this case:
result
In the first loop iteration, bboxes 0, 1, and  3 are removed. This does not make any sense. Bboxes 1 and 3 do not overlap at all. Additionally these bboxes overlap with only bbox 0 which is removed in the same iteration as well. The correct approach is to remove only tested bbox 0.

After this fix, the output is as expected:
result

### DIFF
--- a/ppocr/postprocess/picodet_postprocess.py
+++ b/ppocr/postprocess/picodet_postprocess.py
@@ -292,6 +292,7 @@ class PicoDetPostProcess(object):
                         key=lambda x: x[1]["score"],
                         reverse=True,
                     )[0][0]
-                duplicate_idx.extend([x for x in overlaps if x != keep])
+                if keep != i:
+                    duplicate_idx.append(i)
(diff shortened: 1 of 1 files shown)


leo-q8:
@trueMiskin Please provide the specific commands you used, so I can reproduce and test .

PR #15076, opened 2025-04-27
Fix dead link for diving into OCR

timminator:
I found a dead link in the Readme on the start page of the repository.
I updated it with the correct link.


SWHL (approved):
LGTM

PR #15638, opened 2025-06-08
Fix memory leak when mkldnn is enabled

### LINKED ISSUE #15631: Memory leak when using PaddleOCR 3.0.1/3.0.0 with argument device=cpu
### 🔎 Search before asking

- [x] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.
- [x] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.
- [x] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

### 🐛 Bug (问题描述)

PaddleOCR 3.0.1/3.0.0 has a memory leak when used with the argument command line argument `--device cpu` or in a python script when used with `device="cpu"`. RAM usage increases steadily in this case and does never decrease again. This is especially bad on a folder with a lot of images. On a test folder with around 2000 images I saw a ram usage of about 9 GB!

Image

### 🏃‍♂️ Environment (运行环境)

- OS: Windows 11
- PaddleOCR 3.0.1
- PaddlePaddle 3.0.0 (CPU version)
- 16GB RAM
- Installed via pip in a venv with Python 3.11

### 🌰 Minimal Reproducible Example (最小可复现问题的Demo)

Example in python:

```python

ocr = PaddleOCR(
    use_doc_orientation_classify=False, 
    use_doc_unwarping=False, 
    use_textline_orientation=False,
    device="cpu") # Argument that causes the issue

result = ocr.predict(r"Path\to\folder\with\a\lot\of\images")
for res in result:
    res.print()
    res.save_to_img("output")
    res.save_to_json("output")
```

Example from the command line:

```powershell
paddleocr ocr -i "Path\to\folder\with\a\lot\of\images" --lang ch --use_doc_orientation_classify False --use_doc_unwarping False --use_textline_orientation false --device cpu
```

timminator:
This PR sets a reasonable cache capacity when enabling mkldnn. Previously the cache capacity was set to unlimited, resulting in memory leaks. The chosen cache limit "10" is the same value that was previously used in PaddleOCR 2.10:

https://github.com/PaddlePaddle/PaddleOCR/blob/75526f0a5052a7ea90c012560aadc68f67024157/tools/infer/utility.py#L334-L335

Depends on: https://github.com/PaddlePaddle/PaddleX/pull/4148

Fixes #15631

### DIFF
--- a/paddleocr/_common_args.py
+++ b/paddleocr/_common_args.py
@@ -79,6 +79,8 @@ def prepare_common_init_args(model_name, common_args):
         enable_mkldnn = common_args["enable_mkldnn"]
         if enable_mkldnn:
             pp_option.run_mode = "mkldnn"
+            # cache 10 different shapes for mkldnn to avoid memory leak
+            pp_option.mkldnn_cache_capacity = 10
         else:
             pp_option.run_mode = "paddle"
     pp_option.cpu_threads = common_args["cpu_threads"]


timminator:
This PR was superseded by #15660, so it can be closed.

PR #15863, opened 2025-06-25
Add Mongolian character dictionary

btseee:
Adding Mongolian Cyrillic characters dictionary


cuicheng01:
Hello, this is more like a dictionary of Cyrillic letters, while Mongolian generally refers to traditional Mongolian script, such as ᠮ, ᠣ, ᠩ, etc.

btseee:
Dear @cuicheng01 ,

First Mongolians mostly use extended cyrillic letters such as Ө and Ү.  So most OCR's tend to forget this.
Second, Mongolians primarily use Extended Cyrillic and secondary use Mongolian Script. But that is a good idea i will add that also.

cuicheng01:
> But that is a good idea i will add that also.

Thank you for your contribution. However, I believe that our dictionary should include all characters used in Mongolian contexts, including English characters, to avoid recognition issues in Mongolian scenarios.

Issue #16457, opened 2025-09-11
Segmentation fault on Mac OS


itsmevictor:
### 🔎 Search before asking

- [x] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.
- [x] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.
- [x] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

### 🐛 Bug (问题描述)

I'm getting segmentation fault when using the `ocr.predict` function on any local image. See below for minimal reproducible example. 

Any help would be very appreciated. 

### 🏃‍♂️ Environment (运行环境)

```
Apple M2 Pro
32GB RAM
10 CPUs
Mac OS Sequoia 15.6
paddleocr 3.2.0 
paddlepaddle 3.1.0
paddlex 3.2.1 
```

### 🌰 Minimal Reproducible Example (最小可复现问题的Demo)

```
# Initialize PaddleOCR instance
from paddleocr import PaddleOCR
ocr = PaddleOCR(
    use_doc_orientation_classify=False,
    use_doc_unwarping=False,
    use_textline_orientation=False)

# Run OCR inference on a sample image 
result = ocr.predict(input='image.jpg')

# Visualize the results and save the JSON results
for res in result:
    res.print()
    res.save_to_img("output")
    res.save_to_json("output")
```

where `image.jpg` can be accessed at [this imgur link](https://imgur.com/a/qaP8rEB). Then, running it gives: 

```
❯ python simple_paddle_test.py
/opt/homebrew/lib/python3.11/site-packages/paddle/utils/cpp_extension/extension_utils.py:715: UserWarning: No ccache found. Please be aware that recompiling all source files may be required. You can download and install ccache from: https://github.com/ccache/ccache/blob/master/doc/INSTALL.md
  warnings.warn(warning_message)
Creating model: ('PP-OCRv5_server_det', None)
Model files already exist. Using cached files. To redownload, please delete the directory manually: `/Users/victor/.paddlex/official_models/PP-OCRv5_server_det`.
Creating model: ('PP-OCRv5_server_rec', None)
Model files already exist. Using cached files. To redownload, please delete the directory manually: `/Users/victor/.paddlex/official_models/PP-OCRv5_server_rec`.
[...]

PR #16462, opened 2025-09-11
fix: Prevent SIGBUS on large images by applying safe default resize in OCR pipeline on macOS

scyyh11:
### Summary

* Fixes a **macOS-only crash** (#16457 ) during OCR when processing large images (> 1296×1296).
* Introduces a **platform-guarded default** that caps the long edge at 1280 pixels using a `max` constraint, but only when users have not set their own detection size limits.

---

### Root Cause

* On macOS, forwarding large images directly to the detector causes a **low-level crash in the underlying C/C++ inference stack** during detection preprocessing/inference.
* The Python API passed full-resolution images by default when `text_det_limit_side_len` and `text_det_limit_type` were unset.
* This failure occurs outside Python’s exception handling and results in a hard crash.

---

### Fix

* Add a Darwin-specific safety fallback inside `PaddleOCR.predict_iter`:

  * If both `text_det_limit_side_len` and `text_det_limit_type` are `None`, set:

    * `text_det_limit_side_len = 1280`
    * `text_det_limit_type = "max"`
* File modified:

  * `paddleocr/_pipelines/ocr.py:185`

---

### Impact

* **User experience**: Prevents sudden process termination on macOS when handling large input images.
* **Behavioral change**:

  * Linux/Windows remain unaffected.
  * macOS applies safe defaults **only when user has not provided explicit values**.
* **Backwards compatibility**: Fully preserved — explicit user configs override defaults.

---

### DIFF
--- a/paddleocr/_pipelines/ocr.py
+++ b/paddleocr/_pipelines/ocr.py
@@ -17,6 +17,7 @@
 # maintainability?
 
 import sys
+import platform
 import warnings
 
 from .._utils.cli import (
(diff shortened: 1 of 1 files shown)


scyyh11:
The issue is caused by a Paddlepaddle issue and is fixed in the main repo https://github.com/PaddlePaddle/Paddle/pull/75261 .

PR #17544, opened 2026-01-24
fix docs

zhang-prog:


zhang-prog:
in https://github.com/PaddlePaddle/PaddleOCR/pull/17560

PR #17679, opened 2026-02-09
fix: add langchain compatibility shim for newer versions (fixes #17674)

### LINKED ISSUE #17674: Wrong dependency on old langchain version
### 🔎 Search before asking

- [x] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.
- [x] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.
- [x] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

### 🐛 Bug (问题描述)
依赖一个很老的langchain版本

### 🏃‍♂️ Environment (运行环境)

Image

### 🌰 Minimal Reproducible Example (最小可复现问题的Demo)

```python

# 初始化 OCR（GPU）
ocr = PaddleOCR(
    use_gpu=True,      
    lang="ch",         
    use_angle_cls=True   
)

# 对图片做 OCR
result = ocr.ocr("test.png", cls=True)

# 打印识别文本
for line in result[0]:
    print(line[1][0])
```

Ansarimajid:
### Bug Fix
Fixes PaddleOCR crash caused by LangChain >=0.1.x internal refactor
(`ModuleNotFoundError: langchain.docstore`).

### 🔗 Related Issue
Open Issue #17674

### ❗ Problem
Recent LangChain versions removed or refactored `langchain.docstore`.
PaddleOCR (via `paddlex`) still imports:
`from langchain.docstore.document import Document`,
causing immediate failure during initialization when newer LangChain
versions are installed.

### ✅ Solution
- Introduced a lightweight compatibility layer that:
  - Dynamically provides deprecated LangChain modules when missing
  - Redirects to new module locations when available
  - Gracefully no-ops if LangChain is not installed
- Applied shim at package initialization to ensure compatibility
  before `paddlex` imports occur
- No dependency pinning or API changes introduced

### 🧪 Verification
- Reproduced error with LangChain >=0.1.x
- Verified successful PaddleOCR initialization after fix
- Confirmed no regression in OCR inference

### 📌 Notes
This approach preserves backward compatibility while avoiding strict
dependency pinning, which could otherwise break downstream applications
already using newer LangChain versions.

### DIFF
--- a/paddleocr/_utils/langchain_shim.py
+++ b/paddleocr/_utils/langchain_shim.py
@@ -1,3 +1,17 @@
+# Copyright (c) 2025 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import sys
 import types
 


paddleocr/_utils/langchain_shim.py
```diff
@@ -0,0 +1,53 @@
+import sys
```
cuicheng01, line 15: Here you need to refer to other .py files to add copyright and other information, otherwise the check may not pass

Ansarimajid: I’ve added the license header to match the existing files. Please let me know if anything else is needed. Thanks!

Bobholamovic:
Thank you for your contribution! Your idea is excellent. However, in PaddleOCR 3.x, PaddleX serves as the underlying library of PaddleOCR, so I believe this issue might be more appropriately addressed in PaddleX (since direct users of PaddleX may also encounter it). Perhaps you could refer to [this PR](https://github.com/PaddlePaddle/PaddleX/pull/4981/changes) and submit a corresponding fix to PaddleX? Please remember to target the develop branch instead of the release branch.

Ansarimajid:
> Thank you for your contribution! Your idea is excellent. However, in PaddleOCR 3.x, PaddleX serves as the underlying library of PaddleOCR, so I believe this issue might be more appropriately addressed in PaddleX (since direct users of PaddleX may also encounter it). Perhaps you could refer to [this PR](https://github.com/PaddlePaddle/PaddleX/pull/4981/changes) and submit a corresponding fix to PaddleX? Please remember to target the develop branch instead of the release branch.

Thank you for the guidance!
That makes perfect sense since PaddleOCR 3.x depends on PaddleX internally.

I will prepare a corresponding compatibility fix in PaddleX and submit a PR targeting the develop branch as suggested.

Thanks again for the helpful direction!

Ansarimajid:
> > Thank you for your contribution! Your idea is excellent. However, in PaddleOCR 3.x, PaddleX serves as the underlying library of PaddleOCR, so I believe this issue might be more appropriately addressed in PaddleX (since direct users of PaddleX may also encounter it). Perhaps you could refer to [this PR](https://github.com/PaddlePaddle/PaddleX/pull/4981/changes) and submit a corresponding fix to PaddleX? Please remember to target the develop branch instead of the release branch.
> 
> Thank you for the guidance! That makes perfect sense since PaddleOCR 3.x depends on PaddleX internally.
> 
> I will prepare a corresponding compatibility fix in PaddleX and submit a PR targeting the develop branch as suggested.
> 
> Thanks again for the helpful direction!

@Bobholamovic I’ve opened PaddleX PR (https://github.com/PaddlePaddle/PaddleX/pull/4997) , implementing a global dynamic compatibility solution that supports both older and newer LangChain versions.

PR #17722, opened 2026-02-15
Fix greek dict

IordanisSap:
I am trying to fine-tune for handwritten Greek but some characters are wrong or missing.

The character ´ (U+00B4) is used which is the acute accent. The correct one is ΄ (U+0384) which is the Greek tonos. 

Additionally the Left-Pointing Double Angle Quotation Mark with symbol « was missing (the right one with » already exists)


zhangyubo0722:
Thank you very much for your PR and contribution! Since this change requires model adaptation and retraining, the current model will become unusable. We will handle the necessary model adaptation and subsequent fixes, so this PR will not be merged for now.

Issue #17164, opened 2025-11-23
Issues related to PDF and image-based layout analysis when using PaddleOCR-VL


francisggum:
### 🔎 Search before asking

- [x] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.
- [x] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.
- [x] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

### 🐛 Bug (问题描述)

I am performing inference by deploying with the docker compose provided in the tutorial.
While proceeding with the work, a few issues have occurred, and I would like to get help.

1. There seems to be a problem with rendering text in PDFs.
Among many PDF documents I have, sometimes the text cannot be read. According to the layout analysis results, it seems that text rendering has failed.
In the image below, the left side shows Paddle OCR VL's image layout analysis result, and the right side shows the same page screenshot in a PDF viewer.
Other libraries like pymupdf, pdfplumber, and pdf.js render well.
**Is there any plan to improve the PDF rendering engine or to support third-party rendering engines?**

**

<img width="2024" height="1304" alt="Image" src="https://github.com/user-attachments/assets/3fe5fa1a-fb0b-4fd0-86e8-99f1565aa724" />

**

[평생케어nh종합보장보험(무배당)___2022-04-01___약관.pdf](https://github.com/user-attachments/files/23696602/nh.___2022-04-01___.pdf)

2. To solve the missing text issue by myself, I called the /layout-parsing API using PDF page images instead of the PDF file, and confirmed that it is processed correctly.
However, the /layout-parsing API seems to handle only one image file at a time. So I configured the process by calling the API as many times as the number of pages. But this method is much slower than handling PDF files.
On an RTX 3090, PDF files take about 1.5 seconds per page, whereas image files take about 4 seconds per page. If the default 10-page limit setting is removed for PDF files, it seems to handle up to around 200 pages without issue.
**Is there a way to send multiple image files at once to the /layout-parsing API?**

3. After some consideration, I devised and tested my own method by creating a PDF file filled with page images and calling the /layout-parsing API with the PDF file. 
The missing text issue was resolved, but the speed is similar to calling individual image files sequentially, about 4 seconds per page.
**Is the internal method to handle PDF files different from that for image files?**

[...]

PR #17913, opened 2026-04-12
Make doc preprocessing options configurable in MCP server

kanlac:
## Summary

Add `--use_doc_orientation_classify` and `--use_doc_unwarping` CLI arguments (and corresponding `PADDLEOCR_MCP_USE_DOC_ORIENTATION_CLASSIFY` / `PADDLEOCR_MCP_USE_DOC_UNWARPING` environment variables) to the MCP server.

Previously these options were hardcoded to `False` with no way for users to override them. The default remains `False` to preserve backward compatibility.

## Problem

When processing scanned PDF documents via the MCP server in service mode (`aistudio`, `self_hosted`), layout detection can miss entire text blocks — lines are silently dropped from the output with no error or warning. Enabling `useDocUnwarping` and `useDocOrientationClassify` in the API request resolves the issue.

Since these options are hardcoded to `False` and not exposed as configuration, users have no way to work around this without patching the source code.

Related: #17164

## Usage

```bash
# Via environment variables
PADDLEOCR_MCP_USE_DOC_UNWARPING=true paddleocr_mcp ...

# Via CLI arguments
paddleocr_mcp --use_doc_unwarping --use_doc_orientation_classify ...
```

## Changes

- `mcp_server/paddleocr_mcp/__main__.py`: Add two new CLI arguments with env var fallbacks
- `mcp_server/paddleocr_mcp/pipelines.py`: Replace hardcoded `False` with configurable instance variables

### DIFF
--- a/docs/version3.x/deployment/mcp_server.en.md
+++ b/docs/version3.x/deployment/mcp_server.en.md
@@ -84,8 +84,9 @@ Convert images containing formulas and tables to editable csv/Excel format:
     - [2.3 Working Modes Explained](#23-working-modes-explained)
     - [2.4 Using `uvx`](#24-using-uvx)
 - [3. Running the Server](#3-running-the-server)
-- [4. Parameter Reference](#4-parameter-reference)
-- [5. Known Limitations](#5-known-limitations)
+- [4. Tool Parameter Reference](#4-tool-parameter-reference)
+- [5. Server Parameter Reference](#5-server-parameter-reference)
+- [6. Known Limitations](#6-known-limitations)
 
 ## 1. Installation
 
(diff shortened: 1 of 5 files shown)


Bobholamovic (commented):
Could you also update the documentation accordingly? Also, would it make more sense to let the agent decide whether to apply doc preprocessing, rather than fixing these settings at server init?

kanlac:
Thanks for the suggestion! I updated the PR to expose the document preprocessing options as per-tool-call parameters instead of server initialization settings, so agents can decide whether to enable `use_doc_unwarping` and `use_doc_orientation_classify` for each request.

PR #17924, opened 2026-04-15
fix(ppstructurev3): escape html-sensitive OCR text in table markdown output

jimmyzhuu:
Addresses #16096

## Summary

This PR fixes an HTML escaping issue in `PPStructureV3` table markdown export.

When OCR text inside table cells contains HTML-sensitive content such as `<recv .../>` or `<pause .../>`, the current table HTML assembly path may inject raw OCR text directly into `<td>` nodes. This makes the markdown output render incorrectly and can blur the boundary between original OCR content and generated HTML structure.

This PR narrows the fix to the `PPStructureV3` table export path only.

## Changes

- add a local PaddleOCR patch for PaddleX table post-processing
- escape HTML-sensitive table cell OCR text with `html.escape(..., quote=True)`
- preserve a single outer `<b>...</b>` wrapper so bold formatting is not lost
- apply the patch to both table post-processing modules
- add regression tests for:
  - raw HTML-sensitive OCR text such as `<recv .../>`
  - bold-wrapped content
  - patch application on both post-processing modules

## Scope

This PR does not change `PaddleOCRVL`.
It only fixes the `PPStructureV3` table markdown export path.

## Tests

```bash
pytest tests/pipelines/test_patch_layout_parsing.py tests/pipelines/test_patch_table_markdown.py -q
```

### DIFF
--- a/paddleocr/_pipelines/_patch_table_markdown.py
+++ b/paddleocr/_pipelines/_patch_table_markdown.py
@@ -1,0 +1,125 @@
+# Copyright (c) 2025 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Patches for PP-StructureV3 table markdown escaping.
+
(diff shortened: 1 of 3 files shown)


jimmyzhuu:
已补一个纯格式化提交，处理掉本次 CI 里的 black / pre-commit 改写问题。当前 test-pr 的失败点仍在依赖安装阶段（paddlepaddle 版本解析失败），看起来不是这次表格 markdown 转义改动本身引起的。

Bobholamovic (commented):
It would be better to submit a PR to PaddleX, instead of to PaddleOCR. Please see the comment here: https://github.com/PaddlePaddle/PaddleOCR/commit/efca37e0fbaff10a22fd9cb61ac9735270cfbd2f#r182928994

jimmyzhuu:
Closing this PR because the fix belongs in PaddleX table post-processing rather than a PaddleOCR-side patch. I have prepared the source-level fix in PaddleX and will follow up there instead.

PR #17930, opened 2026-04-16
fix: use safe YAML loaders for config parsing

Hinotoi-agent:
## Summary
This PR hardens PaddleOCR's shared config parsing paths by replacing unsafe YAML deserialization with safe parsing semantics.

The current code uses `yaml.load(..., Loader=yaml.Loader)` on attacker-controlled configuration files and CLI override values. In Python, `yaml.Loader` accepts Python-object tags such as `!!python/object/apply`, which can trigger arbitrary code execution during config parsing.

This patch switches the affected config loaders to `yaml.SafeLoader` and adds regression coverage for the main `tools/program.py` config entry points.

## Security issues covered
| Issue | Impact | Severity |
|-------|--------|----------|
| Unsafe YAML deserialization in config loading | attacker-controlled config files or `-o/--opt` values can trigger Python object construction during parsing | High |

## Before this PR
- shared config helpers used `yaml.load(..., Loader=yaml.Loader)`
- CLI override parsing in `ArgsParser._parse_opt()` evaluated YAML values with the unsafe loader
- loading a malicious YAML file through `tools/program.py` could execute Python-object tags during parsing
- no regression test ensured dangerous YAML tags were rejected in the main shared config path

## After this PR
- shared config helpers use `yaml.SafeLoader`
- CLI override parsing rejects Python-object tags instead of constructing them
- malicious YAML config payloads fail during parsing instead of executing
- regression tests lock in rejection of Python-object tags for the main config and `--opt` path

## Why this matters
PaddleOCR exposes shared configuration-loading helpers that are used by training and export tooling. Those helpers should treat config files and CLI override values as data, not as executable Python object graphs.

Using `yaml.Loader` breaks that boundary. A malicious YAML payload can turn a config parse into code execution. In practice, this matters for anyone who loads untrusted configs, copies training/export commands from third parties, or automates PaddleOCR jobs in CI or shared environments.

## Attack flow
```text
attacker-controlled YAML file or -o value
    -> PaddleOCR config helper calls yaml.load(..., Loader=yaml.Loader)
        -> Python-object tag such as !!python/object/apply is constructed
            -> attacker-controlled code runs during config parsing
```

## Affected code
- `tools/program.py`
- `benchmark/PaddleOCR_DBNet/utils/util.py`
- `test_tipc/supplementary/config.py`
- `configs/rec/multi_language/generate_multi_language_configs.py`

## Root cause
- untrusted config input was parsed with `yaml.Loader` instead of a safe loader
- config parsing treated attacker-controlled YAML as capable of constructing Python objects
- the same unsafe pattern existed in both file-based config loading and CLI override parsing

## CVSS assessment
| Issue | CVSS v3.1 | Vector |
|-------|-----------|--------|
| Unsafe YAML deserialization in config loading | 8.0 High | `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H` |

Rationale:
- attacker control is the supplied config file or CLI override value
- exploitation requires user interaction to load the malicious config
- successful exploitation can execute arbitrary code in the context of the PaddleOCR process

## Safe reproduction steps
1. On vulnerable code, identify the shared config helpers in `tools/program.py`.
2. Confirm that `load_config()` uses `yaml.load(..., Loader=yaml.Loader)` on a supplied YAML file.
3. Confirm that `ArgsParser._parse_opt()` uses `yaml.load(..., Loader=yaml.Loader)` on supplied `-o/--opt` values.
4. Prepare a YAML payload containing a Python-object tag such as `!!python/object/apply`.
5. Load the payload through the config file path or the CLI override path.
6. Observe that vulnerable code evaluates the payload during parsing.
7. On patched code, confirm the same payload is rejected by `yaml.SafeLoader`.

## Expected vulnerable behavior
On vulnerable code, a malicious YAML payload using Python-object tags is accepted by the parser and executed during config loading instead of being rejected as invalid config data.

## Changes in this PR
- replace `yaml.Loader` with `yaml.SafeLoader` in the affected config parsing paths
- harden shared CLI override parsing against Python-object YAML tags
- add a regression test covering `tools/program.py` file-based config loading
- add a regression test covering `tools/program.py` `--opt` parsing

## Files changed
| Category | Files | What changed |
|----------|-------|--------------|
| Shared config hardening | `tools/program.py` | switched config file and CLI override parsing to `yaml.SafeLoader` |
| Related config helpers | `benchmark/PaddleOCR_DBNet/utils/util.py`, `test_tipc/supplementary/config.py`, `configs/rec/multi_language/generate_multi_language_configs.py` | replaced the same unsafe loader pattern in sibling config utilities |
| Regression tests | `tests/test_safe_yaml_config_loading.py` | added tests that malicious Python-object YAML tags are rejected |

## Maintainer impact
This change is intentionally narrow:
- it does not change normal scalar/list/dict YAML parsing behavior
- it does reject Python-specific YAML object tags that should not be accepted from untrusted config input
- it keeps the fix scoped to config parsing and does not change model behavior or inference logic

## Fix rationale
Configuration should be treated as declarative data. `yaml.SafeLoader` enforces that boundary by supporting normal YAML types while rejecting Python-object construction tags.

That is the correct security boundary for PaddleOCR config parsing because config files and CLI overrides should never be able to instantiate arbitrary Python objects during parsing.

## Type of change
- [x] Security fix
- [x] Bug fix
- [ ] New feature
- [ ] Breaking API change
- [ ] Documentation-only change

## Test plan
Commands run:

```bash
cd /Users/lennon/.hermes/hermes-agent && venv/bin/python -m pytest /tmp/PaddleOCR/tests/test_safe_yaml_config_loading.py -q
cd /tmp/PaddleOCR && python3 -m py_compile tools/program.py benchmark/PaddleOCR_DBNet/utils/util.py configs/rec/multi_language/generate_multi_language_configs.py test_tipc/supplementary/config.py tests/test_safe_yaml_config_loading.py
```

Observed result:
- regression tests passed (`2 passed`)
- touched Python files compiled successfully

## Disclosure notes
This PR is intentionally bounded to unsafe YAML deserialization in shared config parsing paths.

It does not claim broader remote exploitability by default. The concrete issue fixed here is that attacker-controlled config input was parsed with an unsafe YAML loader capable of constructing Python objects during parsing.


Hinotoi-agent:
I pushed a follow-up fix for the CI setup on this branch:
- Python 3.8 CI now uses `paddlepaddle==3.0.0` because `3.2.0` is not available for `cp38` on the configured Paddle CPU index
- GPU workflow now installs the released PaddleX dependency range before the editable package install

I also re-ran the relevant local regression test for this PR successfully.

The new GitHub Actions runs were created, but they finished immediately with `conclusion: action_required` and no jobs were started. That looks like a fork-PR workflow approval/policy blocker rather than a remaining code failure.

Could a maintainer please approve/enable the workflow run for this PR so the checks can execute on the updated branch?

.github/workflows/test_gpu.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             python -m pip config set global.index-url https://pypi.tuna.tsinghua.edu.cn/simple
             python -m pip install pytest
             if [ -f requirements.txt ]; then python -m pip install -r requirements.txt; fi
+            python -m pip install 'paddlex[ocr,genai-client,ie,trans]>=3.4.0,<3.5.0'
```
luotao1: The same problem also occurs in https://github.com/PaddlePaddle/PaddleOCR/pull/17931#discussion_r3110901992. Please remove it.

Hinotoi-agent: Removed the redundant PaddleX install from `.github/workflows/test_gpu.yml` and pushed the change in `7750a8a33e`. `python -m pip install -e ".[all]"` already covers that dependency path, so the extra line was unnecessary.

Hinotoi-agent:
I pushed a follow-up for the GPU CI collection issue seen on this branch.

Current head: `a8c8420c46f10749d1965a8dfd64d933a9960841`

Changes:
- replaced direct `sys.modules[...] = ...` test stubs with `monkeypatch.setitem(sys.modules, ...)` in `tests/test_safe_yaml_config_loading.py`
- also registered the dynamically loaded `paddleocr_tools_program` module through `monkeypatch`
- this prevents fake `paddle`, `numpy`, and `ppocr` modules from leaking into later broad GPU test collection

Local validation:
- `pytest -q tests/test_safe_yaml_config_loading.py` — 2 passed
- `pytest -q tests/test_safe_yaml_config_loading.py tests/pipelines/test_patch_layout_parsing.py` — 15 passed
- `uvx black --check tests/test_safe_yaml_config_loading.py` — passed
- `python -m compileall -q tests/test_safe_yaml_config_loading.py` — passed
- `git diff --check` — passed

The new fork workflow runs for this head are currently `action_required`, so they need maintainer approval before GitHub Actions will execute again.

.github/workflows/test_gpu.yml
Hinotoi-agent: Follow-up: I have now restacked this PR onto the current `main`, so the branch no longer carries the old workflow changes at all. Current head: `04c7ec9f7e43de09ba04dcac27ab227ccdf7218c`. The PR is back to the intended safe-YAML change only, plus a focused regression test.

Bobholamovic (dismissed):

Hinotoi-agent:
CI triage update for the current head `a8c8420c46f10749d1965a8dfd64d933a9960841`:

- `PaddleOCR PR Tests` — passed
- `PaddleOCR Code Style Check` — passed
- `PaddleOCR PR Tests GPU` — failed before test collection because the workflow could not clone PaddleX:
  - `fatal: unable to access 'https://github.com/PaddlePaddle/PaddleX.git/': gnutls_handshake() failed: Error in the pull function.`

The later `ModuleNotFoundError: No module named 'paddlex'` collection errors appear downstream of that failed dependency clone, not from this safe-YAML patch. The previous reviewer-requested workflow cleanup is already resolved/outdated; this branch is scoped to the YAML loader change and its regression coverage.

Hinotoi-agent:
Rebased this branch onto current `PaddlePaddle:main` so it includes the upstream CI updates from `#17954`/`#17976` and drops the stale behind-base state.

Current head: `b3e42937e4`

Validation run on the rebased branch:
- `python -m pytest tests/test_safe_yaml_config_loading.py -q` ✅ (2 passed)
- `git diff --check upstream/main...HEAD` ✅

The branch remains scoped to the safe-YAML loader changes plus the focused regression test.

Bobholamovic (approved):

Bobholamovic (approved):

PR #17931, opened 2026-04-16
[security] fix(data): restrict TableMaster LMDB pickle loading

Hinotoi-agent:
## Summary
This PR hardens the TableMaster LMDB dataset-loading boundary by rejecting executable pickle payloads before deserializing dataset metadata or sample records.

- replaces unrestricted `pickle.loads(...)` calls in `LMDBDataSetTableMaster` with a restricted unpickler that only accepts basic built-in container/value types needed by the dataset format
- rejects malicious `__len__` metadata and sample records before attacker-controlled reducers can execute
- preserves expected loading for normal TableMaster LMDB payloads that contain an integer length plus tuple-based sample records
- adds regression tests covering both blocked code-execution payloads and accepted benign dataset payloads

## Security issues covered
| Issue | Impact | Severity |
|-------|--------|----------|
| Unsafe pickle deserialization in `LMDBDataSetTableMaster` LMDB metadata loading | Opening a malicious LMDB dataset can trigger arbitrary code execution during `__len__` deserialization | High |
| Unsafe pickle deserialization in `LMDBDataSetTableMaster` sample loading | Opening a malicious LMDB sample record can trigger arbitrary code execution during per-record deserialization | High |

## Before this PR
- `LMDBDataSetTableMaster.load_hierarchical_lmdb_dataset()` called `pickle.loads(txn.get(b"__len__"))` on attacker-controlled LMDB metadata
- `LMDBDataSetTableMaster.get_lmdb_sample_info()` called `pickle.loads(txn.get(str(index).encode("utf8")))` on attacker-controlled LMDB sample records
- no validation limited pickle contents to the simple scalar/container types the dataset format actually needs
- no regression tests covered malicious LMDB payloads for this TableMaster-specific loader path

## After this PR
- TableMaster LMDB metadata and sample records are deserialized through a restricted unpickler that rejects unsupported globals before code execution can occur
- `__len__` metadata must deserialize to an integer, or loading stops with an explicit error
- malformed or malicious sample payloads are rejected and treated as invalid dataset entries instead of being executed
- regression tests now lock in both the blocked exploit path and the expected benign dataset behavior

## Why this matters
- the broken trust boundary was: attacker-controlled LMDB bytes -> unrestricted pickle deserialization -> code execution during dataset loading
- anyone who can supply or replace a TableMaster LMDB dataset consumed by PaddleOCR can potentially execute code in the permissions context of the training or evaluation process
- this is especially risky for shared datasets, downloaded corpora, CI experiments, or benchmark pipelines where dataset provenance is weaker than code provenance

## Attack flow
```text
attacker-controlled LMDB dataset value
    -> LMDBDataSetTableMaster metadata/sample loader
        -> unrestricted pickle deserialization
            -> attacker-controlled reducer execution
                -> arbitrary code execution in the PaddleOCR process
```

## Affected code
| Issue | Files |
|-------|-------|
| Unsafe LMDB `__len__` deserialization | `ppocr/data/lmdb_dataset.py` |
| Unsafe LMDB sample-record deserialization | `ppocr/data/lmdb_dataset.py` |
| Regression coverage for both paths | `tests/test_safe_lmdb_pickle_loading.py` |

## Root cause
Issue 1: Unsafe LMDB metadata deserialization
- `load_hierarchical_lmdb_dataset()` trusted the raw `__len__` LMDB value and passed it directly to `pickle.loads`
- the trust boundary failed because arbitrary LMDB bytes were treated as safe Python objects instead of untrusted external input

Issue 2: Unsafe LMDB sample-record deserialization
- `get_lmdb_sample_info()` trusted each LMDB record and passed it directly to `pickle.loads`
- the trust boundary failed because the loader accepted executable pickle opcodes even though the format only needs basic built-in data types

## CVSS assessment
| Issue | CVSS v3.1 | Vector |
|-------|-----------|--------|
| Unsafe LMDB metadata deserialization | 8.8 High | `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H` |
| Unsafe LMDB sample-record deserialization | 8.8 High | `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H` |

Rationale:
- both paths require a victim to open or process an attacker-controlled LMDB dataset, but once that happens the pickle reducer executes with the full confidentiality, integrity, and availability impact of the PaddleOCR process

### 1. Unsafe LMDB metadata deserialization
1. Create a malicious pickle payload whose reducer has a harmless local side effect such as writing a marker file.
2. Store that payload in a fake LMDB transaction under the `__len__` key.
3. Call `LMDBDataSetTableMaster.load_hierarchical_lmdb_dataset()` on vulnerable code.
4. Observe that the reducer executes during `pickle.loads(...)` before any type validation.

### 2. Unsafe LMDB sample-record deserialization
1. Create a malicious pickle payload whose reducer has a harmless local side effect such as writing a marker file.
2. Store that payload in a fake LMDB transaction under a sample key such as `b"1"`.
3. Call `LMDBDataSetTableMaster.get_lmdb_sample_info()` on vulnerable code.
4. Observe that the reducer executes during `pickle.loads(...)` before the loader rejects the malformed record.

Prefer safe payloads and bounded local repro steps.

## Expected vulnerable behavior
- attacker-controlled pickle reducers run while PaddleOCR reads TableMaster LMDB metadata or sample records
- vulnerable code executes the reducer before it can reject the payload as malformed or unexpected

## Changes in this PR
- adds `_RestrictedDatasetUnpickler` and `_restricted_pickle_loads()` to constrain TableMaster LMDB deserialization to basic built-in types
- routes both TableMaster LMDB pickle call sites through the restricted loader
- validates that deserialized `__len__` metadata is actually an integer
- rejects malformed or malicious sample payloads unless they deserialize to the expected tuple/list record structure
- adds regression tests for blocked exploit payloads and accepted benign payloads

## Files changed
| Category | Files | What changed |
|----------|-------|--------------|
| Dataset hardening | `ppocr/data/lmdb_dataset.py` | Replaces unrestricted TableMaster pickle deserialization with a restricted unpickler and adds shape/type validation for decoded metadata and records |
| Regression tests | `tests/test_safe_lmdb_pickle_loading.py` | Adds coverage for malicious metadata payloads, malicious sample payloads, and expected benign TableMaster payloads |

## Maintainer impact
- scope is limited to the TableMaster-specific LMDB loader path; unrelated OCR pipelines and unrelated pickle-based utilities are untouched by this patch
- existing benign TableMaster LMDB payloads that use ordinary built-in values remain compatible
- the new helper makes the intended data boundary explicit and easier to preserve during future dataset-loader refactors

## Suggested fix rationale
- this is the right boundary to enforce because LMDB dataset bytes are untrusted input and should not be allowed to instantiate arbitrary Python objects
- the fix is narrow and durable because it constrains only the TableMaster loader path while still allowing the basic value/container types the format expects
- the regression tests are sufficient to prevent reintroduction of unrestricted pickle execution at the two vulnerable TableMaster call sites

## Reference patterns from other software
- secure deserialization patterns generally treat externally supplied state as data, not executable object graphs; this patch follows that principle by allowing only the simple built-in structures the dataset format needs

## Type of change
- [x] Security fix
- [x] Tests
- [ ] Documentation update
- [ ] Refactor with no behavior change

## Test plan
- [x] Regression tests for malicious TableMaster LMDB metadata payloads
- [x] Regression tests for malicious TableMaster LMDB sample payloads
- [x] Compatibility tests for expected benign TableMaster LMDB payloads
- [ ] Full PaddleOCR test suite

Executed with:
- `python -m pytest tests/test_safe_lmdb_pickle_loading.py -q`

If something was not run, state it explicitly and say why.
- the full PaddleOCR test suite was not run because this patch is narrowly scoped to the TableMaster LMDB loader and the targeted regression file directly covers the changed behavior

## Disclosure notes
- this PR is intentionally bounded to the TableMaster LMDB pickle-deserialization path validated in `ppocr/data/lmdb_dataset.py`
- claims here are limited to reviewed code paths and locally reproduced reducer execution using safe marker-file payloads
- no unrelated files were changed


Hinotoi-agent:
I pushed a follow-up fix for the CI setup on this branch:
- Python 3.8 CI now uses `paddlepaddle==3.0.0` because `3.2.0` is not available for `cp38` on the configured Paddle CPU index
- GPU workflow now installs the released PaddleX dependency range before the editable package install
- The touched Python file was formatted to satisfy the code-style job

I also re-ran the LMDB regression test for this PR successfully locally.

The new GitHub Actions runs were created, but they finished immediately with `conclusion: action_required` and no jobs were started. That looks like a fork-PR workflow approval/policy blocker rather than a remaining code failure.

Could a maintainer please approve/enable the workflow run for this PR so the checks can execute on the updated branch?

.github/workflows/test_gpu.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             python -m pip config set global.index-url https://pypi.tuna.tsinghua.edu.cn/simple
             python -m pip install pytest
             if [ -f requirements.txt ]; then python -m pip install -r requirements.txt; fi
+            python -m pip install 'paddlex[ocr,genai-client,ie,trans]>=3.4.0,<3.5.0'
```
Bobholamovic: Why was this line added?

Hinotoi-agent: You’re right — this line was redundant. `python -m pip install -e ".[all]"` already installs the same released PaddleX extra set from `pyproject.toml`, so I removed the extra explicit install to keep this PR scoped to the LMDB fix. Pushed in 4bd246eb2c.

Bobholamovic (dismissed):

Hinotoi-agent:
I restacked this branch onto the current `main` to remove the stale CI/workflow contamination behind the earlier failing checks. The branch now contains only the restricted LMDB pickle-loading fix plus a focused regression test.\n\nCurrent head: `a0a3ae2de5337d89ae43c717aaff08ca7acd885f`\nLocal validation: `pytest -q tests/test_safe_lmdb_pickle_loading.py` ✅ (4 passed)\nPR state: mergeable again\n\nThis force-push also cleared the prior approval on the old head. Could a maintainer please re-enable/approve the workflow run for the new head, and re-approve if the change still looks good?

.github/workflows/test_gpu.yml
Hinotoi-agent: Follow-up: I have now restacked this PR onto the current `main`, so the branch no longer carries the old workflow changes at all. Current head: `a0a3ae2de5337d89ae43c717aaff08ca7acd885f`. The PR is back to the intended LMDB pickle-loading fix only, plus a focused regression test.

Hinotoi-agent:
I pushed a follow-up for the review/GPU collection issue on this branch.

Current head: `982299f8fb2c7825ecae1ff3e686e08b5fe2c83a`

Changes:
- replaced direct synthetic-module writes into `sys.modules` with `monkeypatch.setitem(sys.modules, ...)` in `tests/test_safe_lmdb_pickle_loading.py`
- registered the dynamically loaded `ppocr.data.lmdb_dataset` module through `monkeypatch`
- this keeps the pickle-safety regression tests isolated so their fake modules do not leak into later GPU test collection

Local validation:
- `pytest -q tests/test_safe_lmdb_pickle_loading.py` — 4 passed
- `uvx black --check tests/test_safe_lmdb_pickle_loading.py` — passed
- `python -m compileall -q tests/test_safe_lmdb_pickle_loading.py` — passed
- `git diff --check` — passed

I also checked the latest GPU run after this fix. The remaining GPU failure is from the CI environment failing to clone PaddleX (`GnuTLS recv error (-110)`), which then leaves `paddlex` uninstalled and causes later `ModuleNotFoundError: No module named 'paddlex'` collection errors. That is separate from the test-stub leak fixed here.

Bobholamovic (dismissed):

tests/test_safe_lmdb_pickle_loading.py
```diff
@@ -0,0 +1,123 @@
...
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _stub_module(name, **attrs):
+    module = types.ModuleType(name)
+    for key, value in attrs.items():
+        setattr(module, key, value)
+    sys.modules[name] = module
+    return module
```
scyyh11: This line globally mutates sys.modules without restoring it, which broke the GPU CI.

Hinotoi-agent: Thanks for catching that. Fixed in `982299f8fb` by routing all synthetic module inserts through pytest's `monkeypatch.setitem(sys.modules, ...)`, including the loaded `ppocr.data.lmdb_dataset` module, so the stubs are restored after each test instead of leaking into later GPU test collection.

Validation:
- `pytest -q tests/test_safe_lmdb_pickle_loading.py` — 4 passed
- `uvx black --check tests/test_safe_lmdb_pickle_loading.py` — passed
- `git diff --check` — passed
- added-line secret/sink scan — no findings

The new fork-head workflow run is currently waiting on GitHub's `action_required` approval gate before CI can execute on this commit.

Hinotoi-agent:
CI triage update for the current head `982299f8fb2c7825ecae1ff3e686e08b5fe2c83a`:

- `PaddleOCR PR Tests` — passed
- `PaddleOCR Code Style Check` — passed
- `PaddleOCR PR Tests GPU` — failed before test collection because the workflow could not clone PaddleX:
  - `fatal: unable to access 'https://github.com/PaddlePaddle/PaddleX.git/': GnuTLS recv error (-110): The TLS connection was non-properly terminated.`

The later `ModuleNotFoundError: No module named 'paddlex'` collection errors appear downstream of that failed dependency clone, not from the PR test changes. The review-thread issue about leaked `sys.modules` stubs was fixed in this head by using pytest `monkeypatch.setitem(...)`, and the focused local validation passed:

- `pytest -q tests/test_safe_lmdb_pickle_loading.py` — 4 passed
- `uvx black --check tests/test_safe_lmdb_pickle_loading.py` — passed
- `git diff --check` — passed

Hinotoi-agent:
Rebased this branch onto current `PaddlePaddle:main` so it includes the upstream CI updates from `#17954`/`#17976` and drops the stale behind-base state.

Current head: `eca0c0276b`

Validation run on the rebased branch:
- `python -m pytest tests/test_safe_lmdb_pickle_loading.py -q` ✅ (4 passed)
- `git diff --check upstream/main...HEAD` ✅

The branch remains scoped to the restricted LMDB pickle-loading change plus the focused regression test.

Hinotoi-agent:
Rebased this branch onto the current `PaddlePaddle:main` after the new GPU CI failure on head `eca0c0276b`.

CI triage:
- `test-pr-gpu-impl` failed while installing `paddlex` because the workflow timed out cloning `https://github.com/PaddlePaddle/PaddleX.git`, then later hit `ModuleNotFoundError: No module named 'paddlex'`.
- `test-pr-gpu` failed only because it depends on `test-pr-gpu-impl` succeeding.
- I did not find a PR-specific GPU test failure in the logs; this looks like dependency-clone/transient CI infrastructure, but the PR was also behind `main`, so I refreshed the branch.

Current head: `91ff1ded58eb9a81c639ae84384fc51d89375be7`

Validation on the rebased branch:
- `python -m pytest tests/test_safe_lmdb_pickle_loading.py -q` ✅ (4 passed)
- `git diff --check upstream/main...HEAD` ✅
- `python -m compileall -q ppocr/data/lmdb_dataset.py tests/test_safe_lmdb_pickle_loading.py` ✅
- added-line secret/sink scan over touched files ✅

Notes:
- The configured `uv run ruff check ...` is currently blocked by the repo dependency resolver on this checkout (`paddleocr[all]` requires `paddlex[ocr]>=3.5.0`, while `requires-python >=3.8` conflicts with `safetensors==0.7.0` requiring Python >=3.9). A direct `uvx ruff check` on the touched files still reports the same pre-existing upstream `F841` findings in `ppocr/data/lmdb_dataset.py` (`batch_size`, `raw_name`).
- Immediately after the force-push, GitHub had not yet attached new check runs to the new head and the PR showed `BLOCKED` with an empty check rollup. The previous failed GPU run was on the old head `eca0c0276b`.

Bobholamovic (approved):

PR #17933, opened 2026-04-16
[Feat] Add PP-OCRv6 iOS Demo

Bobholamovic:


timminator:
Hi! Sorry for bothering. I saw the upcoming PRs regarding a PPOCRv6 model but there were no changes made to the general OCR pipeline yet. Will this also get an update or has PPOCRv6 a different use case distinct from the PPOCRv5 model?
Thank you for your time!

Bobholamovic:
> Hi! Sorry for bothering. I saw the upcoming PRs regarding a PPOCRv6 model but there were no changes made to the general OCR pipeline yet. Will this also get an update or has PPOCRv6 a different use case distinct from the PPOCRv5 model? Thank you for your time!

PP-OCRv6 is still under development. It is expected to be used in the same way as PP-OCRv5, namely through the general OCR pipeline.

PR #17936, opened 2026-04-17
Clarify that PaddleOCR-VL manual-install docs describe the verified Python range

### LINKED ISSUE #17935: Clarify whether the PaddleOCR-VL manual-install Python range is a verified range or a hard requirement
## Summary

Several PaddleOCR-VL manual-install docs currently state that Python 3.8–3.13 is a hard requirement.

From a reader's perspective, that wording reads as “newer Python versions will not work”. In practice, that is stricter than what the docs seem to intend.

## Why this is confusing

The same docs are describing a **documented manual-install path**, not a strict runtime gate enforced by PaddleOCR-VL itself.

I locally validated the following before opening this issue:

- editable install of current `paddleocr` on Python 3.13: success
- editable install of current `paddleocr` on Python 3.14: success
- editable install of current `langchain-paddleocr` on Python 3.14: success
- install/import of released `paddleocr==3.4.1` on Python 3.14: success
- `paddleocr --help` on Python 3.14: success

That does **not** prove full official 3.14 support across all environments, but it does suggest the current wording is better framed as a **verified/documented range** rather than a hard prohibition.

## Suggested direction

Would you be open to changing the repeated wording in the PaddleOCR-VL docs from a hard requirement to something like:

> This guide officially covers Python 3.8–3.13; newer versions may still work when compatible PaddlePaddle wheels are available, but they are outside the documented verification scope here.

I can send a small docs-only PR if this direction makes sense.

shaun0927:
## Summary

Clarify the repeated PaddleOCR-VL manual-install wording so it describes the **documented verification range** instead of reading like a hard runtime prohibition.

## Why

The current sentence says Python 3.8–3.13 is required. That is easy to read as “newer Python versions will not work”, while the more conservative and accurate meaning appears to be that 3.8–3.13 is the range covered by the current docs/tests for these guides.

This PR keeps the docs conservative without expanding the official support promise.

## Changes

- update the repeated manual-install sentence in the PaddleOCR-VL main guide
- apply the same clarification to the hardware-specific PaddleOCR-VL guides
- keep scope docs-only

## Validation

- verified all repeated PaddleOCR-VL manual-install statements were updated consistently
- local package validation used for wording justification:
  - editable install/import of current `paddleocr` on Python 3.13 and 3.14 succeeded
  - editable install/import of current `langchain-paddleocr` on Python 3.14 succeeded
  - install/import of released `paddleocr==3.4.1` on Python 3.14 succeeded

Fixes #17935

### DIFF
--- a/docs/version3.x/pipeline_usage/PaddleOCR-VL-AMD-GPU.en.md
+++ b/docs/version3.x/pipeline_usage/PaddleOCR-VL-AMD-GPU.en.md
@@ -48,7 +48,7 @@ If you wish to start the service in an environment without internet access, repl
 
 ### 1.2 Method 2: Manually Install PaddlePaddle and PaddleOCR
 
-If you cannot use Docker, you can also manually install PaddlePaddle and PaddleOCR. Python version 3.9–3.13 is required.
+If you cannot use Docker, you can also manually install PaddlePaddle and PaddleOCR. This guide documents Python 3.9–3.13 as the verified range.
 
 **We strongly recommend installing PaddleOCR-VL in a virtual environment to avoid dependency conflicts.** For example, use the Python venv standard library to create a virtual environment:
 
--- a/docs/version3.x/pipeline_usage/PaddleOCR-VL-AMD-GPU.md
+++ b/docs/version3.x/pipeline_usage/PaddleOCR-VL-AMD-GPU.md
@@ -48,7 +48,7 @@ docker run -it \
 
 ### 1.2 方法二：手动安装 PaddlePaddle 和 PaddleOCR
 
-如果您无法使用 Docker，也可以手动安装 PaddlePaddle 和 PaddleOCR。要求 Python 版本为 3.9–3.13。
+如果您无法使用 Docker，也可以手动安装 PaddlePaddle 和 PaddleOCR。本文档验证过的 Python 版本范围为 3.9–3.13。
 
 **我们强烈推荐您在虚拟环境中安装 PaddleOCR-VL，以避免发生依赖冲突。** 例如，使用 Python venv 标准库创建虚拟环境：
 
--- a/docs/version3.x/pipeline_usage/PaddleOCR-VL-Huawei-Ascend-NPU.en.md
+++ b/docs/version3.x/pipeline_usage/PaddleOCR-VL-Huawei-Ascend-NPU.en.md
@@ -51,7 +51,7 @@ If you wish to start the service in an environment without internet access, repl
 
 ### 1.2 Method 2: Manually Install PaddlePaddle and PaddleOCR
 
-If you cannot use Docker, you can also manually install PaddlePaddle and PaddleOCR. Python version 3.9–3.13 is required.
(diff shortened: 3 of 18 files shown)


Bobholamovic:
Thanks for the contribution! The intent behind this PR — softening the wording from a hard requirement to a documented verification range — is a reasonable direction, and helps align reader expectations with what the docs actually guarantee.

That said, I'd like to raise a concern about this part of the new wording:

> "newer versions may still work when compatible PaddlePaddle wheels are available"

This phrasing ties whether PaddleOCR works on a newer Python version to the availability of a compatible PaddlePaddle wheel, and I think this is problematic for two reasons:

1. **PaddleOCR's supported Python versions are not strictly coupled with those of the PaddlePaddle framework.** PaddleOCR has many dependencies beyond PaddlePaddle itself (numpy, opencv, scipy, etc.), each with its own Python version support range. In addition, the upcoming PaddleOCR 3.5.0 release will introduce support for switching inference backends, after which Paddle won't necessarily be the only backend in play.

2. **The phrasing implicitly makes an overly optimistic assumption** — that if the PaddlePaddle framework supports a given Python version, PaddleOCR will too. In practice, a compatible PaddlePaddle wheel is neither a sufficient nor a necessary condition for PaddleOCR to work on that Python version.

Additionally, the local validation in the PR description only covers successful `import`, which isn't really enough to support "newer versions may still work" — the PaddleOCR-VL pipeline has a long dependency chain, and a successful import doesn't imply the pipeline is fully functional.

I'd suggest removing or rewriting the "when compatible PaddlePaddle wheels are available" clause, and keeping only the message that 3.8–3.13 is the documented verification range — without implying anything about what happens outside that range. This would actually be more aligned with the "keep scope conservative, don't expand support claims" intent stated in your PR description. Thanks again for the contribution!

shaun0927:
Thanks for the thorough review, @Bobholamovic — both of your points make sense, especially the reminder that (a) PaddleOCR's Python support is not 1:1 with PaddlePaddle's (numpy/opencv/scipy, and the upcoming 3.5.0 backend switch), and (b) a successful `import` is not a functional signal for the pipeline.

I've dropped the "newer versions may still work when compatible PaddlePaddle wheels are available…" clause entirely and kept only the verified-range statement, with no implication about versions outside that range. Updated wording (applied consistently across all 16 EN + ZH guides):

- EN: `This guide documents Python 3.8–3.13 as the verified range.`
- ZH: `本文档验证过的 Python 版本范围为 3.8–3.13。`

Pushed as b781b27. Let me know if the new wording reads clearly for the docs' intent — happy to refine further.

Bobholamovic (dismissed):

Bobholamovic:
Please resolve the conflicts

shaun0927:
Rebased onto current main, bumped the numbers to 3.9–3.13 to follow #17954, and kept the verified-range wording you'd approved earlier. Also squashed the two earlier commits into one to keep the diff small. Let me know if anything else needs tweaking.

luotao1 (approved):

Issue #18320, opened 2026-08-20
[PP-OCRv6][Performance] High CPU inference latency and low CPU utilization - need Small model accuracy with lower latency


Nandhu910:
### 🔎 Search before asking

- [x] I have searched the PaddleOCR [Docs](https://paddlepaddle.github.io/PaddleOCR/) and found no similar bug report.
- [x] I have searched the PaddleOCR [Issues](https://github.com/PaddlePaddle/PaddleOCR/issues) and found no similar bug report.
- [x] I have searched the PaddleOCR [Discussions](https://github.com/PaddlePaddle/PaddleOCR/discussions) and found no similar bug report.

## Problem

We are trying to achieve low OCR latency while maintaining good
accuracy for invoices and table documents.

We initially ran PaddleOCR with:

2 vCPU / 4 GB RAM

and then increased the infrastructure to:

4 vCPU / 8 GB RAM

However, increasing the resources provided only around 0.2 seconds
latency improvement.

More importantly, during OCR inference the container does not appear
to fully utilize the available CPU.

Although approximately 3.5 vCPU is allocated to the pod, observed CPU
consumption is typically only around 1-1.5 vCPU.

Memory consumption is also only approximately 1-1.5 GB.

Therefore, increasing the pod from 2 vCPU to approximately 4 vCPU has
not resulted in proportional inference performance improvement.

Example profiling:

OCR_INFERENCE_MS=7549
OCR_PARSING_MS=25
TOTAL_REQUEST_MS=8115

OCR_INFERENCE_MS=6981
OCR_PARSING_MS=14
TOTAL_REQUEST_MS=7017

OCR_INFERENCE_MS=8871
OCR_PARSING_MS=42
TOTAL_REQUEST_MS=8937

OCR_INFERENCE_MS=8555
OCR_PARSING_MS=20
TOTAL_REQUEST_MS=9300

Almost all request latency is inside PaddleOCR inference.
Application parsing takes only a few milliseconds.

## Environment

We are running PaddleOCR as an OCR API on AWS EKS.

- PaddleOCR: 3.7.0
- PaddlePaddle: 3.2.0
- PaddleX: 3.7.x
- Python: 3.10
- Device: CPU
- AWS Region: eu-north-1
- EC2 instance: c6i.xlarge
- Instance resources: 4 vCPU / 8 GiB RAM
- Pod CPU request/limit: 3500m
- Pod memory: 5-6 GiB
- Architecture: linux/amd64

[...]

PR #18375, opened 2026-09-22
docs: document CPU OCR tuning options

dongyikuan919:
## Summary

- document explicit CPU thread and stage batch-size tuning for OCR deployments
- explain that larger batches need workload-specific benchmarking
- call out the PP-OCRv6 Tiny/Small accuracy trade-off for narrow invoice text

This complements #18320 by making the available CPU tuning knobs discoverable in both Chinese and English quick-start docs.

## Verification

- Documentation-only change; no runtime code changed.


dongyikuan919:
I have read the CLA Document and I hereby sign the CLA

